PR #16, opened 2019-11-25
use list extend instead of directly assigning

krnick:
as title

need to same with next_list2


krnick:
Thank you for your reminding, it exactly as you said.

PR #30, opened 2020-01-06
add unit test for xrule

krnick:
add test for find_previous_method method


sidra-asa:
I think this commit could be merged into dev branch first, and will be merged into master with next new feature or code refactory. @18z @krnick  How do you think ?

krnick:
@sidra-asa got it, I will complete this unit test of xrule.py in dev branch until new feature.

PR #31, opened 2020-01-06
fix typo in tests/Object/test_apkinfo.py

krnick:


sidra-asa:
Thanks for the typo fix. It fixed my mistake.

PR #32, opened 2020-01-07
Ready to release stable v20.01

krnick:
1. add more bytecode instruction into pyeval.py.

2. add dunder method `__repr__` to each object to better represent the state of object.

3. add unit test for XRule.py.

4. follow sphinx docstring style to generate doc automatically in the future.

5. using list.clear() to clean the result from the previous rule to avoid duplicate results in detail report.

6. add logging module to pyeval.py to keep track of the bytecode usage.

7. update version number from v19.10 to v20.01

### DIFF
--- a/quark/Evaluator/pyeval.py
+++ b/quark/Evaluator/pyeval.py
@@ -12,6 +12,7 @@ class PyEval:
             "invoke-direct": self.INVOKE_DIRECT,
             "invoke-static": self.INVOKE_STATIC,
             "move-result-object": self.MOVE_RESULT_OBJECT,
+            "move-result-wide": self.MOVE_RESULT_WIDE,
             "move-result": self.MOVE_RESULT,
             "new-instance": self.NEW_INSTANCE,
             "const-string": self.CONST_STRING,
@@ -100,6 +101,24 @@ class PyEval:
             pre_ret = self.ret_stack.pop()
             variable_object = VarabileObject(reg, pre_ret)
             self.table_obj.insert(index, variable_object)
+        except Exception as e:
+            # No element in pop
+            pass
+
+    def MOVE_RESULT_WIDE(self, instruction):
+        """
+        move-result-wide vx
+
+        Move the long/double result value of the previous method invocation into vx,vx+1.
+        """
+        reg = instruction[1]
+        index = int(reg[1:])
+        try:
+            pre_ret = self.ret_stack.pop()
+            variable_object = VarabileObject(reg, pre_ret)
+            variable_object2 = VarabileObject(f"v{index + 1}", pre_ret)
+            self.table_obj.insert(index, variable_object)
+            self.table_obj.insert(index + 1, variable_object2)
         except Exception as e:
             # No element in pop
             pass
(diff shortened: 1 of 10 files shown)


quark/Evaluator/pyeval.py
```diff
@@ -1,32 +1,65 @@
...
+
+import logging
+from datetime import datetime
+
 from quark.Objects.TableObject import TableObject
 from quark.Objects.VarabileObject import VarabileObject
 
 MAX_REG_COUNT = 40
+TIMESTAMPS = datetime.now().strftime('%Y-%m-%d-%H-%M')
+LOG_FILENAME = f"{TIMESTAMPS}.quark.log"
```
sidra-asa: Is it necessary to rotate the log by minutes?

krnick: It will only be executed once, for example ("2020-01-08-13-29.quark.log") to log which command has executed when the quark instruction is executed, or maybe I misunderstand this usage?

sidra-asa: Apparently, it will create a new log file every execution. Since our engine is command line tool, not a running service. It is OK to separate each runtime log if this is what we want. How do you think ?

quark/Evaluator/pyeval.py
```diff
@@ -1,32 +1,65 @@
...
+            "const-wide/high16": self.CONST_WIDE_HIGHSIXTEEN,
+            # array
             "aget-object": self.AGET_OBJECT,
         }
 
         self.table_obj = TableObject(MAX_REG_COUNT)
         self.ret_stack = []
 
-    def __invoke(self, instruction):
+    def _invoke(self, instruction):
```
sidra-asa: At line 70 - 84, I think it can become only one loop.

quark/Objects/VarabileObject.py
```diff
@@ -21,72 +18,83 @@ def __init__(self, register_name, value, called_by_func=None):
...
+        Setter of value.
 
         :param value:
-        :return:
+        :return: None
         """
         self._value = value
 
-    def get_all(self):
+    def get_current_status(self):
```
sidra-asa: The method is similar with __repr__. Is it necessary ?

krnick: Same here, I will remove this function.

quark/Evaluator/pyeval.py
krnick: Thanks for your comments!

quark/Evaluator/pyeval.py
krnick: I see. Any ideas on how to modify it?

sidra-asa: Actually, I prefer to rotate the log file by day, not by minute.

krnick: got it!, I will change it by day!

quark/Evaluator/pyeval.py
krnick: @sidra-asa  When I tried to simplify this code, it leads to the unit test failed. I would leave this part unchanged until we start refactoring later, I still try to find the reason.

sidra-asa: I see. Leave this to next refactory

sidra-asa:
LGTM

krnick:
@sidra-asa thank you for your reviewing!! 😄

PR #41, opened 2020-02-07
Added quark_inside to the docs

krnick:
This is the manual for the quark-engine implementation principle. The purpose is to make it easier for everyone to understand how quark-engine is implemented.

Because there is a lot of content, this pull request will be merged to master directly. If there are problems with the content, we will fix it step by step in the future.

@18z 
@sidra-asa


krnick:
Added quark_inside to the docs.

PR #42, opened 2020-02-12
update description of rules and weighted score

krnick:


sidra-asa:
@krnick I don't understand the modification with yscore. Can you explain the yscore criterion?

PR #47, opened 2020-03-13
fix typo 'RuleoOject' to 'RuleObject'

pulorsok:


krnick:
LGTM. @pulorsok , thank you for the fix! 👍

Issue #46, opened 2020-03-12
RecursionError: maximum recursion depth exceeded while calling a Python object


krnick:
It seems that the problem of recursive maximum still exists in find_previous_method and upperfunc.

```
Traceback (most recent call last):
  File "/Users/nick/Desktop/quark-engine/quark/Objects/xrule.py", line 71, in find_previous_method
    item, top_method, pre_method_list)
  File "/Users/nick/Desktop/quark-engine/quark/Objects/xrule.py", line 71, in find_previous_method
    item, top_method, pre_method_list)
  File "/Users/nick/Desktop/quark-engine/quark/Objects/xrule.py", line 71, in find_previous_method
    item, top_method, pre_method_list)
  [Previous line repeated 989 more times]
  File "/Users/nick/Desktop/quark-engine/quark/Objects/xrule.py", line 59, in find_previous_method
    method_set = self.apkinfo.upperfunc(class_name, method_name)
  File "/Users/nick/Desktop/quark-engine/quark/Objects/apkinfo.py", line 60, in upperfunc
    method_set = self.find_method(class_name, method_name)
  File "/Users/nick/Desktop/quark-engine/quark/Objects/apkinfo.py", line 44, in find_method
    if len(list(result)) > 0:
  File "/Users/nick/.local/share/virtualenvs/quark-engine-NJwlz3ey/src/androguard/androguard/core/analysis/analysis.py", line 1776, in find_methods
    classname = bytes(mutf8.MUTF8String.from_str(classname))
  File "/Users/nick/.local/share/virtualenvs/quark-engine-NJwlz3ey/src/androguard/androguard/core/mutf8.py", line 102, in from_str
    c = cls(encode(s))
  File "/Users/nick/.local/share/virtualenvs/quark-engine-NJwlz3ey/src/androguard/androguard/core/mutf8.py", line 64, in encode
    ord_array = [i for i in map(lambda x: ord(x), s)]
  File "/Users/nick/.local/share/virtualenvs/quark-engine-NJwlz3ey/src/androguard/androguard/core/mutf8.py", line 64, in <listcomp>
    ord_array = [i for i in map(lambda x: ord(x), s)]
  File "/Users/nick/.local/share/virtualenvs/quark-engine-NJwlz3ey/src/androguard/androguard/core/mutf8.py", line 64, in <lambda>
    ord_array = [i for i in map(lambda x: ord(x), s)]
RecursionError: maximum recursion depth exceeded while calling a Python object
```

[...]

PR #48, opened 2020-03-19
Xrule bugfix

sidra-asa:
Bugfix in xrule.py #46


krnick:
Thanks @sidra-asa  for fixing issue # 46

PR #49, opened 2020-03-20
Bugfix for issue #46

krnick:


krnick:
It's time to fix issue #46

Related references

https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

https://docs.python-guide.org/writing/gotchas/

PR #51, opened 2020-03-20
fixes bug issue in #50

### LINKED ISSUE #50: Stage Checking Bug
1. if the first stage check fails, it will continue to next stage.

Right [Here](https://github.com/quark-engine/quark-engine/blob/d2783684fe463c6b6f101cde22fae034b9260a8a/quark/Objects/xrule.py#L231-L232).

```python=
        if set(rule_obj.x1_permission).issubset(set(self.apkinfo.permissions)):
            rule_obj.check_item[0] = True
```

2. The second stage of the inspection should check the first or second function, as long as one of them appears to meet the standard.

Right [here](https://github.com/quark-engine/quark-engine/blob/d2783684fe463c6b6f101cde22fae034b9260a8a/quark/Objects/xrule.py#L234-L238).

krnick:
1. fixes #50.
2. Reduce the nested if conditionals to improve readability.
3. Replace the string concatenation with f-string.

### DIFF
--- a/quark/Objects/xrule.py
+++ b/quark/Objects/xrule.py
@@ -45,6 +45,8 @@ class XRule:
         # Sum of the each rule
         self.score_sum = 0
 
+        self.level_2_reuslt = []
+
     def find_previous_method(self, base_method, top_method, pre_method_list, visited_methods=None):
         """
         Find the previous method based on base method before top method.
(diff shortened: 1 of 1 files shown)


18z:
> 3. Replace the string concatenation with f-string.
to boost the speed?

krnick:
> > 1. Replace the string concatenation with f-string.
> 
> To boost up the speed?

@18z Yes

18z:
> > > 1. Replace the string concatenation with f-string.
> > 
> > 
> > To boost up the speed?
> 
> @18z Yes

Ni how bang !!!

krnick:
@18z  @sidra-asa thanks you all, you are how ban ban.

PR #52, opened 2020-03-23
Porting androguard to quark-engine

krnick:
Porting androguard version 3.4 to quark-engine project to prevent androguard from no longer being maintained.

In the past using androguard, we all have to rely on `pip install androguard` from Github, but there is a problem, if something goes wrong with androguard, quark-engine might crash.

But in fact, we only need the decompile function of androguard, so I ported this function to our project.

In addition to improving the stability of quark-engine, it also increases the speed of pipenv installation.


sidra-asa:
If there's new release of Androguard, or new enhancement to decompiling functions in the future. For our engine, what will we do?

18z:
This issue is worth concerning! 

Some pros and cons came to my mind.

Pros:
- No need to worry about the maintain of androguard.

Cons:
- The code base will be too large.
- We only use one or two modules in androguard, no need to include the whole project.
- How do you make sure our porting version is up to date when there's update or bug fix of the androguard.

To solve this problem, I suggest that we do a deep dive study of androguard. And just extract the exact modules we need. Also, we need to fully understand how those module work and understand each line of the code. After the study, we need to document everything down for future maintainers.

krnick:
In my opinion, it is worth doing that so I will keep this pull request open until we have done for this work, which can be our record to achieve this goal.

However, we don't need to fully understand the whole project of androguard since we can use this branch to remove useless code, and also use code debugger to trace the execution flow in androguard while using quark.

PS: this androguard branch is clone by the day before yesterday.

krnick:
The below link is my code trace of androguard. In this GitHub branch, I have separated the code we must use and deleted the unused code in androguard.

https://hackmd.io/-eyqx_5wQ2-r3kyM_jI3Tg?both

I also used the package called [`vulture`](https://pypi.org/project/vulture/), which can help us to identify the unused code so that we can delete it.

In addition, the analysis speed has also been improved. The following image shows `time` command three times for analysis.






However, merging this branch would significantly reduce our code coverage. I don’t want to include the test code in androguard, because I think androguard has not been updated for a long time.

Still, even if androguard is updated, I think I have the ability to modify it directly from the source code.

krnick:
More fast after removing the decompiler code!

krnick:
So far, the following image is the modules we need, of course, it can reduce the code even more.



(91.3 MB on disk) of 787 items (androguard porject now)
(5.4 MB on disk) of 63 items (what we really use in androguard now)

krnick:
## Conclusion

I prefer to keep this pull request open until  Androguard is no longer maintained one day because Androguard is still got maintained now.

Secondly, we may have to spend a lot of effort to understand the entire content of Androguard, and even though that I had reduced it to 60 files in Androguard, but only nearly a second faster than before.

Third, our test coverage will be significantly reduced and increase the complexity of the project.

Finally, this branch has extracted the functions we need, so it can be used when we needed in the future.

krnick:
> If there's new release of Androguard, or new enhancement to decompiling functions in the future. For our engine, what will we do?

After careful analysis, I found that we did not use the decompile function. For now, we only use the parser of APK files and the function of cross-references.

Therefore, I decided to move the above function only into our project from androguard.

> Cons:
> 
> * The code base will be too large.

Yes, it does, but I would move the cross-reference and the apk parser functions only. Also, write the test and documentation for them.

> * We only use one or two modules in androguard, no need to include the whole project.

That's right, by removing useless modules, we can greatly reduce code base and increase execution speed.

> * How do you make sure our porting version is up to date when there's an update or bug fix of the androguard.

I don’t plan to rely on androguard anymore, based on the functions we use in androguard, I will make our own module to obtain the function same as the androguard, which means that there is no longer need to use androguard in the future.

In addition, the function of androguard 3.4 is what we need. However, it has been stuck at 3.3 for a long time, which makes it impossible to use pip install directly. Instead, we need to rely on GitHub URL to install the master version from GitHub, which makes it impossible for us to install Quark on PyPi.

sidra-asa:
Great works. 
Since we will build our own cross-reference functions and APK parser, what is next for us ?

krnick:
@sidra-asa 

As I see it, I will open another branch to make APK Parser and cross-reference, and then merge into master when it is finished.

Could you please help me with this part of the test and the code review.

Thanks!

sidra-asa:
Sure.

> @sidra-asa
> 
> As I see it, I will open another branch to make APK Parser and cross-reference, and then merge into master when it is finished.
> 
> Could you please help me with this part of the test and the code review.
> 
> Thanks!

PR #55, opened 2020-04-21
add pre-commit to identify simple issue before you start code review

krnick:
[pre-commit](https://pre-commit.com/) are useful for identifying simple issues before submission to code review.

If there is something in the code that does not match to the coding style, it will prohibit you from using the commit and help you correct it before using it again.

Only when all the check are passed will you be allowed to use commit command.

You can also directly execute `pre-commit run --all-files` to correct the code


18z:
Nice work yo!

quark/Evaluator/pyeval.py
```diff
@@ -11,18 +11,19 @@
 
 MAX_REG_COUNT = 40
 TIMESTAMPS = datetime.now().strftime('%Y-%m-%d')
-LOG_FILENAME = f"{TIMESTAMPS}.quark.log"
+LOG_FILENAME = f'{TIMESTAMPS}.quark.log'
```
sidra-asa, line 14: According to our doc about coding style, the strings is delimited by double quotes (“). 
[link](https://quark-engine.readthedocs.io/en/latest/coding_style.html#strings)

krnick: @sidra-asa thanks for reminding, I have removed the double quotes fixer

PR #84, opened 2020-11-02
Create Requirements.txt

yashomer1994:
/*
This file will contains Dependencies packages which needs to be installed on the host system before installing quark-engine to avoid dependencies errors. */

 pip install androguard =>3.4.0a1
pip install pipenv
pip install django

### DIFF
--- a/Requirements.txt
+++ b/Requirements.txt
@@ -1,0 +1,6 @@
+/*
+This file will contains Dependencies packages which needs to be installed on the host system before installing quark-engine to avoid dependencies errors. */
+
+pip install androguard =>3.4.0a1
+pip install pipenv
+pip install django 


18z:
> /*
> This file will contains Dependencies packages which needs to be installed on the host system before installing quark-engine to avoid dependencies errors. */
> 
> pip install androguard =>3.4.0a1
> pip install pipenv
> pip install django

hi @yashomer1994 

Thank you for the PR.
I have few questions before you keep developing this pr.  :slightly_smiling_face:

1. We already use pipenv as our package manager. Why do you still need to install androguard with pip?
2. Why do you need to install django?

yashomer1994:
Hi, 
Thanks for reviewing the Pr 

i am sorry may be there might have issue as this is my first pr. 

the pr i have submitted is for macOS . 
1) I have tested on macOS  using pipenv , but it fails it is searching for androguard so its better to install dependency before running setup.py to install. 
2.) Django is my idea to make web analysis as an option.

krnick:
Hello @yashomer1994 , 

 Adding Requirements.txt is very useful for some users who like to use it, but since we are already using Pipfile, it is more suitable for development.

Also, your Requirements.txt is not a standard format for requirements files, please check the following link:

https://pip.pypa.io/en/stable/user_guide/#requirements-files

The python package to be installed should mainly be the package currently in use, rather than the unused package for installation, to avoid unnecessary installation time.

Issue #86, opened 2020-11-15
Saving report to json


pokorny-martin:
quark -a to_test\com.fastemulator.gba.apk -r quark\rules --detail -o to_test\quark_gameboy.json

Traceback (most recent call last):
  File "C:\Users\Martin\.virtualenvs\quark-engine-xwtWkN4W\Scripts\quark-script.py", line 33, in <module>
    sys.exit(load_entry_point('quark-engine', 'console_scripts', 'quark')())
  File "C:\Users\Martin\.virtualenvs\quark-engine-xwtWkN4W\lib\site-packages\click\core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "C:\Users\Martin\.virtualenvs\quark-engine-xwtWkN4W\lib\site-packages\click\core.py", line 782, in main
    rv = self.invoke(ctx)
  File "C:\Users\Martin\.virtualenvs\quark-engine-xwtWkN4W\lib\site-packages\click\core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "C:\Users\Martin\.virtualenvs\quark-engine-xwtWkN4W\lib\site-packages\click\core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "d:\desktop\android testing\quark-engine\quark\cli.py", line 128, in entry_point
    data.generate_json_report(rule_checker)
  File "d:\desktop\android testing\quark-engine\quark\Objects\quark.py", line 330, in generate_json_report
    for class_name, method_name in self.quark_analysis.level_2_result:
TypeError: cannot unpack non-iterable MethodAnalysis object

18z:
Thanks for reporting this issue. We'll take a look.

sidra-asa:
@pokorny-martin 
We've done a fix. Would you try with the newer commit and tell us  if any error ?

pokorny-martin:
Works like a charm, thank you.

PR #87, opened 2020-11-16
Bugfix in saving json format report

sidra-asa:
Bugfix to #86, also with testcase with json report generator.
@krnick please take review.


krnick (approved):
LGTM, thank you @sidra-asa for the bug fix and the testcase, it works perfectly.

krnick:
Fixed issue #86, thanks for @sidra-asa good work.

PR #106, opened 2021-01-21
add quark-engine requirements

Surendrajat:
Quark failed to install on a system with Python 3.6 installed. It also failed to run without `git`.


krnick:
Thank you @Surendrajat  for providing the installation requirements of quark, so that users can understand the usage of quark more quickly.

PR #123, opened 2021-02-15
multi-line string fix

lordlabuckdas:
The regex string for URL validation was probably aimed to be implemented as a **multi-line string**, but it was missing parenthesis/brackets and thus the strings following the initial one (for HTTP check) were ignored.


krnick:
Thank you @lordlabuckdas  for fixing this problem, it work like a charm 😄

PR #126, opened 2021-02-15
Improvements in regular expressions with IP address

techathena:


krnick:
Hello @techathena, the regular expression you provided **r"(?:[0-5]{1,3}\.)+(?:[0-5]{1,3})"** will not match when the number is greater than 5, some IP addresses such as 172.0.0.1, 66.66.66.66 will fail with that.

Maybe you can refer to Django's implementation of this [part](https://github.com/django/django/blob/45bef6706ae7436b5c2c503018b21ff70b2e6728/django/core/validators.py#L69)

Also, you can make PR directly to the master branch since the test branch is behind the master for so long. I will update it later.

techathena:
Oh,okay.i should have thought of that.(maybe had too much coffee)
Thanks :)

krnick:
No problem @techathena !

Would you like to improving our previous regular expressions into regular expressions based on Django implementation?

```python
r"(?:25[0-5]|2[0-4]\d|[0-1]?\d?\d)(?:\.(?:25[0-5]|2[0-4]\d|[0-1]?\d?\d)){3}"
```

This should work better and please rebase your commit

Thank you for your kind help 😄

techathena:
On it .thanks  :)

Issue #119, opened 2021-02-09
Quark rules classification outputs twice the same rule in some cases


cryptax:
How to reproduce:

- Get this [malicious sample](https://koodous.com/apks/a03ffa955d9b273cc22acae93dbad2ec8c023d7621b256ea93506587506e3eb3)
- Get Quark v21.02.1
- `quark -a sample.apk -r ./quark-rules -s -c`

You see below crimes for `Lgoogle/com/sjhi/client/zjService;e ` are duplicated.

![quark-stutters](https://user-images.githubusercontent.com/2493690/107402188-fdad5080-6b03-11eb-843d-f8deb28e4461.png)

This is not the case for each function. The bug does not happen all the time.

18z:
@cryptax  We will take a look! Thanks for reporting! 😄

krnick:
Resolved it with PR #137
Fixed by @haeter525 , nice work 👍

PR #137, opened 2021-02-18
Remove duplicated crime description

haeter525:
Quark output crime description twice or more due to the duplication of parent functions. (#119)

Crimes with different bytecodes but the same parent will be output respectively.
Thus, I wrap descriptions with a set in order to filter out the duplicated ones.

### DIFF
--- a/quark/utils/output.py
+++ b/quark/utils/output.py
@@ -27,7 +27,7 @@ def output_parent_function_table(call_graph_analysis_list):
 
         count = 1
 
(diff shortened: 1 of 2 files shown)


krnick:
Thanks to @haeter525 for solving this issue!
It seems that this feature has not been tested. Would you like to adding some unit tests for that?

krnick:
Great job @haeter525, LGTM 😄 
Thanks for adding the unit tests, it works very well!

Issue #118, opened 2021-02-09
Quark engine does not report the same risk based on threshold


cryptax:
I am using [this malicious sample again](https://koodous.com/apks/a03ffa955d9b273cc22acae93dbad2ec8c023d7621b256ea93506587506e3eb3).

I run : `quark -a a03ffa955d9b273cc22acae93dbad2ec8c023d7621b256ea93506587506e3eb3.apk -r ./quark-rules -s` and get a **Moderate Risk**.

![quark-moderate-risk](https://user-images.githubusercontent.com/2493690/107364767-2d486280-6adc-11eb-8109-e1187178f45f.png)

Now, if I set threshold to 100%, I get a **High Risk** !! This does not make sense. The sample did not change, the risk level should be the same whatever threshold I use.

![quark-high-risk](https://user-images.githubusercontent.com/2493690/107364944-697bc300-6adc-11eb-8de9-d55ff7e1cc40.png)

If I try with various thresholds:
- not specified, 20%, 40% : Moderate Risk
- 60%, 80%, 100%: High Risk

Test conditions:
- Quark v21.02.1
- Linux

krnick:
It seems that if a threshold is added when calculating the score, other rules that are not met will be ignored. Thank you @cryptax for reporting this issue.

18z:
@cryptax thanks for reporting this issue. 😄 

Actually, despite that we have designed the so called "malware scoring system".

The scoring system takes effect only when we have enough amounts of detection rules.
And the problem here is, we are still thinking about what amount of rules should be enough.

Now let's skip the problem above and assume that we have enough rules.
Then we have issues concerning adjusting the weighting scores of each rule. :-D

TBH, Quark is now 50% of its power and I believe it will be 100% when we make scoring system really works.

PR #139, opened 2021-02-20
Fix reporting apk risk depends on user-given thershold.

haeter525:
Fix issue #118 .

Extract calculations for sums of tatal scores/weighted scores to a function `update_sum`.
And the function are invoked before each return statement of  `run` method in quark.py.
Finally, add a simple test for it :)

### DIFF
--- a/quark/Objects/quark.py
+++ b/quark/Objects/quark.py
@@ -215,6 +215,15 @@ class Quark:
         :param rule_obj: the instance of the RuleObject.
         :return: None
         """
+
+        def update_sum():
+            weight = rule_obj.get_score(rule_obj.check_item.count(True))
+            score = rule_obj.yscore
+            # add the weight
+            self.quark_analysis.weight_sum += weight
+            # add the score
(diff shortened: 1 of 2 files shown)


krnick:
Hi @haeter525 , thank you for your kind help in fixing this issue!

I think there is no need to write much code. Just move the [code](https://github.com/quark-engine/quark-engine/blob/3d55dc0a3bdfb04b0f8b5ec9e6510b77bb4bf40b/quark/Objects/quark.py#L395-L398) below:

```python
# add the weight
self.quark_analysis.weight_sum += weight
# add the score
self.quark_analysis.score_sum += score
```
where the score and weight are calculated to the bottom of the [show_summary_report](https://github.com/quark-engine/quark-engine/blob/3d55dc0a3bdfb04b0f8b5ec9e6510b77bb4bf40b/quark/Objects/quark.py#L400) function.

haeter525:
I think you are right. It's a better way to solve this issue, which keeps as a small modification as it needs. I will submit another PR later!

PR #140, opened 2021-02-20
Fix score-sum

haeter525:
Fix issue #118 .

By @krnick 's advice, move calculations for sums of tatal scores/weighted scores to `show_summary_report` method.
This calculation should not be blocked by thershold.
Thank @krnick a lot!


krnick:
Nice work @haeter525 , thanks a lot for fixing 😄

PR #150, opened 2021-03-26
Update cli.py

1llusionary:
I don't think execute the freshquark operation when running cli is a good idea, which will cause the quark to run slowly. It's up to the user to decide when to execute the freshquark operation

### DIFF
--- a/quark/Evaluator/pyeval.py
+++ b/quark/Evaluator/pyeval.py
@@ -22,7 +22,8 @@ LOG_FILENAME = f"{config.HOME_DIR}{TIMESTAMPS}.quark.log"
 log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
 handler = FileHandler(LOG_FILENAME, mode='w')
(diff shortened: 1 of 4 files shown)


krnick:
Hello @1llusionary ,

Thanks for your PR, we did consider removing the freshquark when running CLI, could you explain why you want to modify the logger module in this Pull Request?

1llusionary:
basicConfig sets the configuration of root logger, log= logging.getLogger (__ name__ ）create another logger object and inherited from the root logger. When using the API to call quark, if the user creates another logger, the user's log will also be recorded in QUARK_LOG_FILE.

krnick:
Hi @1llusionary ,

Thank you for your detailed and clear explanation, could you reopen another Pull Request for fixing this logger configuration.

We have updated the latest version to "v21.3.4", where we have removed the freshquark in the CLI.

1llusionary:
@krnick 
OK

PR #162, opened 2021-04-26
fix bug in aget-object in quark/Evaluator/pyeval.py

ciastron:
fix bug in `AGET_OBJECT` method in `quark/Evaluator/pyeval.py`
This method takes as parameter `instruction`, 
```
print(instruction)
> ['aget-object', 'v9', 'v12', '11']
```
* If register number (v12 in this case) has a certain value (i.e. an integer of 2 digits), with the original code it is taken the wrong register number, as in this case the expected value should be 12, but it is taken 1.
* Index of array (third parameter of instruction) is an integer, so it makes sense to take "all" the value of `instruction[3]` and not `instruction[3][1]`

The edited lines of code permit to get the correct number of register (12, not anymore 1) and the correct index (11 not anymore 1)

Hope this could help :-)

### DIFF
--- a/quark/Evaluator/pyeval.py
+++ b/quark/Evaluator/pyeval.py
@@ -7,6 +7,7 @@
 # http://pallergabor.uw.hu/androidblog/dalvik_opcodes.html
 
 import logging
+import re
 from datetime import datetime
 
 from quark.Objects.registerobject import RegisterObject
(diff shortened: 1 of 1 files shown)


quark/Evaluator/pyeval.py
```diff
@@ -333,10 +333,10 @@ def AGET_OBJECT(self, instruction):
         try:
 
             array_obj = self.table_obj.get_obj_list(
-                int(instruction[2][1]),
```
ciastron, line 336: if `instrucion[2] = v12` then `instruction[2][1] = 1` but it should be correct 12

ciastron, line 336: if `instrucion[2] = v12` then `instruction[2][1] = 1` but it should be correct 12

ciastron, line 336: if `instrucion[2] = v12` then `instruction[2][1] = 1` but it should be correct 12

krnick (approved):
Nice work @ciastron :tada:, thanks for the bug fixing for the bytecode.
I will handle the unit test now!

PR #163, opened 2021-04-26
Added check to skip permissions check if DEX file

Dil3mm3:
Fix issue with DEX files. 

If the file that has to be analyzed is a DEX file, it does not contain a manifest, so it is useless to check for permissions. In fact, this check will only leads to an unwanted behavior: the permission check will fail (even if the permissions are actually requested) and the analysis on that rule will stop (even if the rule is matched for the 100%)

I just edited two lines of code in the `Quark.py` file to skip permissions check if we are analyzing a DEX file.

### DIFF
--- a/quark/Objects/quark.py
+++ b/quark/Objects/quark.py
@@ -219,7 +219,9 @@ class Quark:
         self.quark_analysis.crime_description = rule_obj.crime
 
         # Level 1: Permission Check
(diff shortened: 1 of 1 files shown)


krnick (approved):
LGTM, thank you @Dil3mm3 for the patch! 🎉

PR #165, opened 2021-05-15
Creation of option to print Report based on Label Rules

ciastron:
With the following `pull request` we (me, @cryptax, @Dil3mm3 and @3aglew0) propose you to add another option to print a report based on *labels* specified inside a rule.

We have noticed they are not used and it could be interesting to print a short report taking into consideration these values. Here an example of output where it is printed for each label (found inside the rules) a description (see explanation below), the number of rules where this label is contained and other detailes described better below.

example_of_output

This option permits to print a report based on label with two different levels of details
1. `quark -a malware_to_be_analysed.apk -r rule_dir -l max`
     print the maximum score for each label (as image above), this would permit us to understand in which topic (represented by label) a malware is more aggressive. For example, looking at the previous output we can see the malware performs with success malicious action related to location, calllog and sms. 
2. `quark -a malware_to_be_analysed.apk -r rule_dir -l detailed`
    print a detail report with all the previous information plus: 
    * Number of rules (with that label) which have a score >= 80%
    * Average score and standard deviation (computed over the all the scores obtained by that specific label). Interesting considerations could be the following:  label with high average and low standard deviation would allow us to say the malware performs a series of malicious actions (with success); then, a high standard deviation means there are some rules which take high score so the malware performs with success only some actions with that label; finally, a low standard deviation and a low average on a certain label means the malware is not performing malicious action on that topic. Example of output:
    
output_detailed_report

The column description allows to add a short and representative sentence about a label, for example for the `callog` the relative description is `Retrieve or manipulate sensitive data from call log`. In order to implement a flexible solution we have thought to add a *csv file* in the same directory of rules with the following structure `label,description`. We have chosen `csv` extension because it is easy to manipulate and it wasn't possible to use a json format since in that folder all json files are interpreted as rules. If this file is not present or a `label,description` pair is absent, the corresponding cell in the label report is filled with `-`. Example of output

output_with_desc

I leave here a sample of the csv file to be put in the folder of the rules ([label_desc.csv](https://github.com/quark-engine/quark-engine/files/6485320/label_desc.csv))

Do not hesitate to contact me for any type of clarification

### DIFF
--- a/quark/Objects/analysis.py
+++ b/quark/Objects/analysis.py
@@ -14,11 +14,19 @@ def init_pretty_table():
     tb.align = "l"
     return tb
 
+def init_label_report_table():
+    # Pretty Table Output
+    tb = PrettyTable()
+    tb.field_names = ["Label", "Description", "Number of rules", "MAX Score %"]
+    tb.align = "l"
+    tb.sortby = "Number of rules"
+    tb.reversesort = True
+    return tb
 
 class QuarkAnalysis:
     __slots__ = ["crime_description", "first_api", "second_api", "level_1_result", "level_2_result", "level_3_result",
                  "level_4_result", "level_5_result", "json_report", "weight_sum", "score_sum", "summary_report_table",
-                 "call_graph_analysis_list", "parent_wrapper_mapping"]
+                 "label_report_table", "call_graph_analysis_list", "parent_wrapper_mapping"]
 
     def __init__(self):
         self.crime_description = ""
@@ -37,6 +45,8 @@ class QuarkAnalysis:
         # Sum of the each rule
         self.score_sum = 0
         self.summary_report_table = init_pretty_table()
+        # label report
+        self.label_report_table = init_label_report_table()
         # Call graph analysis
         self.call_graph_analysis_list = []
 
--- a/quark/Objects/quark.py
+++ b/quark/Objects/quark.py
@@ -18,6 +18,8 @@ from quark.utils.graph import call_graph
 from quark.utils.out import print_info, print_success
 from quark.utils.output import output_parent_function_table, output_parent_function_json
 from quark.utils.weight import Weight
+import pandas as pd
+import os
 
 MAX_SEARCH_LAYER = 3
 CHECK_LIST = "".join(["\t[" + "\u2713" + "]"])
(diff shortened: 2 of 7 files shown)


18z:
Hi @ciastron 

Nice work! I'm all for it!
This feature provides another angle for examining the malware.

krnick:
Thank you all @ciastron @cryptax, @Dil3mm3 and @3aglew0 for making this 😄 , it looks great!

That is a useful and awesome feature to add to the Quark report. I will spend some time reviewing this code.

18z:
BTW, @krnick has a good idea of turning 

118356087-ddae8280-b573-11eb-842e-f562bf21c652

into a radare chart
radar-chart-1-what

krnick (approved):
Hello @ciastron ,

I have reviewed all the code and passed all the tests, so now it works like a charm.

One thing I would like to discuss with you is that the term "Score" used in the label report means the confidence level in the Quark report.

They are confidence 0%, 20%, 40%, 60%, 80%, 100%, which means the latter the stage the more we’re sure that the malicious behavior is practiced.

On the other hand, the "Score" in the Quark report is used in the scoring system. Users can give different weights to different behaviors by themself.

Therefore, replacing "Score" with "Confidence" may be more consistent with our summary report and detail report.



Thank you all for making this awesome feature 😄 .

krnick:
Also, I have uploaded the label_desc.csv to repo [quark-rules](https://github.com/quark-engine/quark-rules/pull/8)

krnick:
> Yes, indeed, rename "Score" by "Confidence".

That's great! @ciastron, would you like to work on it?

ciastron:
Hi @krnick , sorry for my late answer. I am pleased to work on it. According to your note I will change name of variables and string which could lead to confusion with the score you really mean, I will introduce variables and string as Confidence, AVG Confidence...
I close this merge request so I can create a new one with the correct names :smiley:

krnick:
> Hi @krnick , sorry for my late answer. I am pleased to work on it. According to your note I will change name of variables and string which could lead to confusion with the score you really mean, I will introduce variables and string as Confidence, AVG Confidence...
> I close this merge request so I can create a new one with the correct names 😃

No problem! Thank you @ciastron :DD

PR #167, opened 2021-05-22
Choose one rule or label to be analyzed in summary or detailed output + more Rules info in output

Dil3mm3:
With the following pull request we (me, @cryptax, @ciastron and @3aglew0) propose you to add the possibility of choosing a rule or a label to filter the output of quark. Using Quark during the last weeks we thought that having the possibility to choose only a certain rule/label when printing the summary or detailed output can be useful. With the changes we made, we filter the rules that Quark is using to analyze the current APK/DEX file; in this way we are not just applying a filter to the output, but we are also preventing Quark from wasting time analyzing rules on which we are not interested in. 

In order to do so, we changed the `summary` and `detail` options: in few words, we added the possibility to add a parameter after invoking those options. With _Click 8.0.1_ in fact, it is possible to specify options in a way that they can accept zero or more parameters, using the "_flag_value_" attribute.

Thanks to this new feature of Click, now it is possible to invoke the two options in three different ways:

1. `Quark -a sample.apk -s` --> will analyze and print the output of all the rules inside the default directory
2. `Quark -a sample.apk -s label` --> will analyze and print only the output related to the specified label
3. `Quark -a sample.apk -s rule.json` --> will analyze and print only the output related to the specified rule

The same thing happens when using the `-d` option. 

Obviously this is something that can be extended to more rules or labels at the same time if you find it useful.

Moreover, we added the `Filename`column to the `summary` table. We thought that having the possibility to understand immediately which rule is obtaining a certain level of confidence can be useful, for example, to edit them faster or simply to easily identify them inside the rules directory (that can result quite boring reading only the crime associated to that rule). In addition, this is something that is already shown in the detailed output, that's why we thought that it can be useful to have it also in the summary table.
In the same way, in the detailed output only the Rulepath was printed while the rule crime was not shown, for this reason we added it too.

I will show you some output examples here:
summary_label_censored
summary_rule_censored
detail_rule_censored

### DIFF
--- a/quark/Objects/analysis.py
+++ b/quark/Objects/analysis.py
@@ -10,7 +10,7 @@ from prettytable import PrettyTable
 def init_pretty_table():
     # Pretty Table Output
     tb = PrettyTable()
-    tb.field_names = ["Rule", "Confidence", "Score", "Weight"]
+    tb.field_names = ["Filename", "Rule", "Confidence", "Score", "Weight"]
     tb.align = "l"
     return tb
 
--- a/quark/Objects/quark.py
+++ b/quark/Objects/quark.py
@@ -385,9 +385,10 @@ class Quark:
         # add the score
         self.quark_analysis.score_sum += score
 
-    def add_table_row(self, rule_obj, confidence, score, weight):
+    def add_table_row(self, name, rule_obj, confidence, score, weight):
 
         self.quark_analysis.summary_report_table.add_row([
+            name,
             green(rule_obj.crime),
             yellow(confidence),
             score,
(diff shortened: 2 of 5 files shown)


Dil3mm3:
I noticed that there are some issues with PEP8 (missed some whitespaces and similar stuff). I am sorry about that. Of course, if you will find those new features useful, I can close the request and open another one without all these errors.

krnick:
Thank you all (@Dil3mm3, @cryptax, @ciastron, and @3aglew0) for this pull request. Your team really helped Quark a lot!

I second that such a feature is quite helpful. It can greatly expand the usage of Quark and use Quark more efficiently for specific rules/labels you would like to review.

For pep8 issues, you can use the "black" python tool to solve it based on this commit, or you can open another pull request, both of which are great.

Dil3mm3:
Thank you @krnick!

Ok so I will close this issue and open a new one without errors :smile:

PR #170, opened 2021-05-29
Update setup.py to latest click version

Dil3mm3:
Hi all,

Testing Quark we noticed that, without forcing Click to the latest version (8.0.1) in the Setup.py file, the attribute `flag_value`, inserted to let the options `-s` and `-d` take zero or more arguments, is not working. For this reason, I just changed the Setup.py and Click version. Everything should work correctly now :smile:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,7 +30,7 @@ setuptools.setup(
         "androguard==3.4.0a1",
         "tqdm",
         "colorama",
-        "click",
+        "click==8.0.1",
         "graphviz",
         "pandas", 
     ],


krnick:
Hello @Dil3mm3,

Thank you for pointing out this problem!

I have tested `pip3 install click`, `pipenv install click`, and `pip3 install quark-engine`, the default click version of these three commands is 8.0.1, which is now the latest version.

Please help me to check whether your current installation of click is the same. If so, we don’t need to specify the version number.

https://pypi.org/project/pip/#history

Thank you 😄

18z:
Hi @Dil3mm3 

Thanks for the PR! 

I think it's important to lock the version of modules we use in the setup.py.

And I suggest the policy of version locking is: 
1. We tested the latest version first and make sure the latest version works and is stable as the previous version. 
2. And then, we bump the version to the latest stable one. 

Thanks again for reminding us!

18z:
We do have some other friends encounter this problem! Thanks @Dil3mm3 for fixing this up! :D

18z (approved):
Thanks! :D

krnick (approved):
I misunderstood that the latest version will be installed by default, but later found out that this is not the case. Thank you @Dil3mm3 for solving this problem.

Dil3mm3:
You're welcome! 😄

PR #176, opened 2021-06-05
Updated help and README for summary and detail

Dil3mm3:
Hi all,

I made just some small changes that can maybe help users to understand how to properly use the `summary` and `detail` options.
In particular, I updated:

-  The help, trying to explain the possible usage of the `-s `and `-d` options
quark_help
- The README.md, to specify how the two options can be used

I hope these clarifications can be appreciated :smile:

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,17 @@ and the report will look like:
 
 <img src="https://i.imgur.com/g28N7qk.png"/>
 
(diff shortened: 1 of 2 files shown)


krnick:
Thanks for clarifying that, really appreciate it. 😄

Issue #191, opened 2021-06-29
Show Parent Functions' Cross-References in Rule Classification


haeter525:
**Is your feature request related to a problem? Please describe.**

The current rule classification doesn't consider references between parent functions. If a parent function calls another to perform a behavior, it is difficult to check this behavior in the rule classification report.

**Describe the solution you’d like**

I propose a new feature that lists cross-references between parent functions in the rule classification report.
The feature adds extra descriptions to the Crime Description field to show what functions are called by the parent function.

Here is the pattern.

```
* Call {A function called by the parent function}
```

---

Imagine a scenario.

+ `Parent Function 2` does `Behavior 2`.
+ `Parent Function 1` does `Behavior 1` and calls `Parent Function 2`.

For `Parent Function 1`, the current rule classification table below lists one description in the Crime Description field.

```
+-------------------+------------------------------+
| Parent Function   | Parent Function 1            |
+-------------------+------------------------------+
| Crime Description | * Description for Behavior 1 |
+-------------------+------------------------------+
```

The feature adds a line (Line 5) to indicate its reference to `Parent Function 2`.

```
+-------------------+------------------------------+
| Parent Function   | Parent Function 1            |
+-------------------+------------------------------+
| Crime Description | * Description for Behavior 1 |
|                   | * Call Parent Function 2     | <--Add_by_this_feature.
+-------------------+------------------------------+
```

The feature also applies to the JSON version of the rule classification report.

```json
{
    "parent": "Parent Function 1",
    "crime": [
        "Behavior 1",
        "Call Parent Function 2"       # <--Add_by_this_feature.
    ]
}
```

**How can one use this feature?**

According to the additional descriptions, users can refer to the report of the called parent function. Take `Parent Function 1` as an example. Here is the report of  `Parent Function 2`.

```
+-------------------+----------------------------+
| Parent Function   | Parent Function 2          |
+-------------------+----------------------------+
| Crime Description | * Behavior 2 Description   |
+-------------------+----------------------------+
```
Now, the user knows that Parent Function 1 does two behaviors instead of one. With this feature, the user can see how the parent functions interact with each other.

**Will all referenced functions be listed in the field?**

No, only those functions that do at least one behavior are visible.

haeter525:
Hi @krnick ,

Can you kindly take a look at this feature? If you think it is OK, please assign it to me.

Thank you!

krnick:
Closed with #192

PR #192, opened 2021-07-01
Show Parent Functions' Cross-References In Rule Classification

haeter525:
**Description**

The PR adds a new feature that lists references of the parent functions in the rule classification report. (See the text with a yellow underline bellow.)

With this text, the user can see how parent functions interact with each other.

For more details, please refer to issue #191.

**Code Changes**

1. Add a new function to prepare the data of a rule classification report.
2. Add code to search for cross-references.

**Test Plan**

1. Add tests with test data containing cross-references.

### DIFF
--- a/quark/Objects/quark.py
+++ b/quark/Objects/quark.py
@@ -575,8 +575,12 @@ class Quark:
 
     def show_rule_classification(self):
         print_info("Rules Classification")
-        output_parent_function_table(self.quark_analysis.call_graph_analysis_list)
-        output_parent_function_json(self.quark_analysis.call_graph_analysis_list)
+        output_parent_function_table(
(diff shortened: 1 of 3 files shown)


krnick:
I think the way of this graph is not easy for users to understand. I would prefer it can be presented in the following way:



you could achieve it by using the `graphviz` package.

haeter525:
Hi @krnick ,

In the above commits, I added the reference graph according to your advice.
But to avoid the graph over-width, I have to rotate it by 90 degrees.

Now, the generated graph is as the following.


Please take a look. Thank you.

krnick (approved):
Nice work @haeter525 ,
Thanks for this awesome feature!

PR #194, opened 2021-07-05
Add A Standard Interface for Accessing APK Infomation

haeter525:
**Description**
This PR adds a standard interface for Quark to access an APK.

It is the first step to make the analysis code independent of the core library, Androguard.

This PR provides:
1. BaseApkinfo class, a standard interface that defines how Quark interacts with its core library.
2. MethodObject class, used by the standard interface. It is a replacement for MethodAnalysis class.

**Why is there an additional class, MethodObject ?**
Quark uses MethodAnalysis objects as parameters in most of its files. It's not practical to eliminate the dependency file by file.

A working solution I used here is to replace MethodAnalysis with a Quark-owned class.
The only change to the existing code is adjusting the return values in apkinfo.py. And this change will arrive in the next few PRs.

**Code Changes**
+ Add quark/Object/methodobject.py
+ Add quark/Object/baseapkinfo.py

**Test Plans**
+ Add five tests for methodobject.py

### DIFF
--- a/quark/Objects/baseapkinfo.py
+++ b/quark/Objects/baseapkinfo.py
@@ -1,0 +1,183 @@
+import hashlib
+import os.path
+from abc import abstractmethod
+from os import PathLike
+from typing import Dict, List, Optional, Set, Union
+
+from quark.Objects.bytecodeobject import BytecodeObject
+from quark.Objects.methodobject import MethodObject
+
+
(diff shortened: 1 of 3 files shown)


krnick (approved):
Thanks for the PR, LGTM!
Nick work @haeter525

PR #197, opened 2021-07-06
Clean Up the Direct Use of MethodAnalysis

haeter525:
**Description**

This PR deprecates all direct calls to MethodAnalysis's method in Quark.
It ensures that all method information used by Quark is from Apkinfo.

It is the second part of making the analysis code independent of the core library, Androguard. (see the first one [here](https://github.com/quark-engine/quark-engine/pull/194  ))

**Code Changes**

- Add a function lowerfunc in apkinfo.py to get method references.
- Modify the following three files to use lowerfunc instead of get_xref_to.
   1. quark.py
   2. graph.py
   3. output.py

**Test Plans**

- Add a test for lowerfunc in test_apkinfo.py.

### DIFF
--- a/quark/Objects/apkinfo.py
+++ b/quark/Objects/apkinfo.py
@@ -170,6 +170,18 @@ class Apkinfo:
         """
         return {call for _, call, _ in method_analysis.get_xref_from()}
 
+    @functools.lru_cache()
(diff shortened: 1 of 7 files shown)


krnick (approved):
Thanks for the PR, LGTM 👍 
Nick work @haeter525

PR #209, opened 2021-07-10
Update travis.yml to set up Rizin

haeter525:
**Descriptions**
This PR makes Travis CI set up the latest release of Rizin when it starts a test.

Since there is no official Rizin package on `apt`,
it uses a series of commands to download it via [the Github release page](https://github.com/rizinorg/rizin/releases).

Here are the details:
1. Query the file link to the latest version of Rizin via the release page.
2. Use it to download the latest binaries.
3. Unzip the binaries into /opt/rizin and add the executable rizin to the system variable PATH.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,13 @@ language: python
 python: 3.8
 
 before-install:
+  - RIZIN_LATEST=https://github.com/rizinorg/rizin/releases/latest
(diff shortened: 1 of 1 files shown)


krnick (approved):
Thanks for fixing the CI, nice work! @haeter525

Issue #127, opened 2021-02-17
Make Quark's rules support API parameter


krnick:
For now, Quark can only support API and permission detection, but some key points rely on the usage of the parameters. For example, `contentResolver.query` can get the SMS content as well as the contact information, it will get the corresponding information based on the given parameters.

That would be great to add an additional feature to enable Quark to determine which parameters the API uses.

codewithvicky:
Can you Explain in Detail so i can work on this issue sir please

krnick:
Closed in PR #212.

PR #212, opened 2021-07-11
Add an Optional Parameter Filter For JSON Rules

haeter525:
**Description**

This PR proposes to add an optional keyword filter for JSON Rules.
Users can use the keyword filter to match APIs with specified parameters.

**When will the detection be triggered**?

The detection is applied when a rule's confidence reaches 100%.
If no parameters in the corresponding API contain the specified keyword, the rule will be discarded.

**How to use it**

Append key `match_keywords` with a keyword list to an API.
For example, add the keyword `content://call_log/calls` to the second one as below.

```json
{
    "api": [
        {
            "descriptor": "()Landroid/content/ContentResolver;",
            "class": "Landroid/content/Context;",
            "method": "getContentResolver"
        },
        {
            "descriptor": "(Landroid/net/Uri; [Ljava/lang/String; Ljava/lang/String; [Ljava/lang/String; Ljava/lang/String;)Landroid/database/Cursor;",
            "class": "Landroid/content/ContentResolver;",
            "method": "query",
            "match_keywords": [					<--- Added Keyword filter.
                "content://call_log/calls"		<--- The keyword
            ]
        }
    ],
}
```

**Code Changes**
1. `quark/Objects/quark.py`
   + Add a method to check the provided keywords.

**Test Plans**
1. Add four tests for the


krnick:
Thank you @haeter525 for adding this feature!
Resolve related issue in #127.
Awesome!

krnick:
For the `match_keywords`, I would prefer to change it to `keyword`, which is simpler.

haeter525:
Hi @krnick 

I submitted a commit to rename the key. Please take a look.
Thank you.

PR #219, opened 2021-07-19
Update README.md instruction from pip to pip3

sharteeya:
**Description**
- Change install instruction in README.md from `pip` to `pip3`


pulorsok:
Thank you @sharteeya  for the PR!

PR #224, opened 2021-08-04
Fix threshold filtering in the detailed mode.

### LINKED ISSUE #199: Threshold filtering does not work in detailed mode
The command `quark -d -a virus-locker.apk -t 60` outputs details for all rules including those with confidence lesser than 60%. The command line acts as if `-t 60` wasn't specified, whereas this option does work for `-s` (summary) mode. If you issue `quark -s -a virus-locker.apk -t 60`, you do only get the results for rules with confidence >= 60%, which is what I expected.

I think the parsing of options needs to be fixed so that `-t` also works with `-d` and produces detailed output only for rules >=60%. Currently, there is fortunately a workaround by specifying `-d rulenumber.json`, but it is not very convenient if you have several rules.

This is only a _minor issue_, but probably not too difficult to fix :)

**Desktop (please complete the following information):**
 - OS: Linux Mint 20.1
 - Quark 21.6.2
 - Python 3.8.5

haeter525:
Close #199.

### DIFF
--- a/quark/cli.py
+++ b/quark/cli.py
@@ -277,6 +277,7 @@ def entry_point(
 
     # Show detail report
     if detail:
+        threshold_number = int(threshold) if threshold else 0
 
         if detail == "all_rules":
             label_flag = False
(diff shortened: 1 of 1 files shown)


krnick (approved):
Thanks @haeter525 for fixing issue #199 , LGTM!

Issue #225, opened 2021-08-05
issue with quark log and IntelOwl


mlodic:
Hi everyone,

while developing on IntelOwl I have just noticed that when importing 
```
from quark.report import Report
```
from the [Quark analyzer](https://github.com/intelowlproject/IntelOwl/blob/develop-2/api_app/analyzers_manager/file_analyzers/quark_engine.py#L7), in the main directory of the project a log file is created. But it seems that it is never used by our integrations (`<date>.quark.log`)

This [line](https://github.com/quark-engine/quark-engine/blob/master/quark/evaluator/pyeval.py#L20) generates this issue. A proper library should not create files for logging in this way: this should happen only when executing Quark on its own.

Could you please help us by creating that log file exclusively when calling Quark directly?

This is the stack of calls that happens
```
  File "/opt/deploy/intel_owl/./api_app/analyzers_manager/file_analyzers/quark_engine.py", line 7, in <module>
    from quark.report import Report
  File "/usr/local/lib/python3.8/site-packages/quark/report.py", line 7, in <module>
    from quark.core.quark import Quark
  File "/usr/local/lib/python3.8/site-packages/quark/core/quark.py", line 13, in <module>
    from quark.evaluator.pyeval import PyEval
```

Thank you in advance and keep up the great work.

sharteeya:
Hi mlodic,
We'll discuss this issue. Thanks for your report!

krnick:
Thank you @mlodic for letting us know about this issue. We have excluded the Log module in the Quark API and will release the latest version as soon as possible :D

PR #230, opened 2021-08-09
Avoid logging messages in the Dalvik Bytecode Loader

haeter525:
**Description**
Refer to #225 . 
The logging behavior in pyeval.py is now disabled by default. Developers can set a variable in config.py to bring it back.

**Code changes**
+ pyeval.py
  + Apply an if-statement to the logger.
+ config.py
  + Add a variable named DEBUG to indicate whether the logger should run.


krnick (approved):
Thanks @haeter525 for fixing this problem!
LGTM!

PR #231, opened 2021-08-17
Cache commonly used attributes in the Rizin core library

haeter525:
**Description**

To eliminate the repetitive calculation in the Rizin core library (rzapkinfo.py),
this PR caches the return values of the commonly used attribute, `all_methods`, `permissions`, and `class_hierarchy`.

**Code changes**
+ quark/core/rzapkinfo.py
   + Replace decorator, `property` with the cached decorator, `cached_property` in `all_methods`, `permissions`, and `class_hierarchy`.


krnick:
Thanks @haeter525  for the patch! LGTM

PR #232, opened 2021-08-17
Simplify the generator value check in the Androguard core library

haeter525:
**Description**

This PR uses the built-in function to check the generator value.
The new check can eliminate the repetitive calls in the Androguard core library.

**Code changes**
+ quark/core/apkinfo.py
  + Apply the built-in function, [next()](https://docs.python.org/3/library/functions.html#next), to the return value of Androguard API, `AnalyzeAPK.find_methods`.
  + Remove the repetitive calls to Androguard API, `AnalyzeAPK.find_methods`.


krnick:
Nice work! @haeter525

Issue #158, opened 2021-04-17
Methods from the "Extended Class"


ciastron:
Hi all, my friends (@Dil3mm3 and @3aglew0) and I are working to implement new quark rules (version 21.3.2) for a university semester project (our supervisor is @cryptax). We were analyzing Brazking malware (hash SHA256 be3d8500df167b9aaf21c5f76df61c466808b8fdf60e4a7da8d6057d476282b6, let us know if you want the sample).

In a nutshell the problem we have encountered is the following: *we noticed Quark is not able to detect all the API called from an object of a class which extends a noticed Android Class. The root of cause comes from the signature of the API that in the smali code appears with the name of the child class.*

To explain better this problem, we provide the following example: Acessibilidade class is a custom Brazking class which extends Android AccesibilityService class.

```
package com.gservice.autobot;

...

public class Acessibilidade extends AccessibilityService{
    ...
}
```

Acessibilidade is widely used by this malware to perform accessibilty service actions, as the one below:

```
public void Clicar_Pos(int i, int i2) {
        Acessibilidade acessibilidade = this.Contexto;
        if (acessibilidade != null) {
            try {
                Clica(i, i2, acessibilidade.getRootInActiveWindow(), 0);
            } catch (Exception unused) {
            }
        }
    }
```

The incriminant line of code is the one where it is called the method `getRootInActionWindow` which in the smali code appears as following

`invoke-virtual {v0}, Lcom/gservice/autobot/Acessibilidade;->getRootInActiveWindow()Landroid/view/accessibility/AccessibilityNodeInfo;`

In the Context of Quark Rules it makes sense link this API with the PerformAction API of AccessibilityNodeInfo class

```
public void Clica(int i, int i2, AccessibilityNodeInfo accessibilityNodeInfo, int i3) {

        ...
    
                accessibilityNodeInfo.performAction(16);
        
        ...
    
    }
```

Finally we have created the following rule (note that the first API is written with the signature of AccessibilityService class and not Acessibilidade since it would not have sense make a rule so specific for a single malware)

```
{
    "crime": "Use accessibility service to perform action getting root in active window",
    "permission": [],
    "api": [
        {
            "class": "Landroid/accessibilityservice/AccessibilityService;",
            "method": "getRootInActiveWindow",
            "descriptor": "()Landroid/view/accessibility/AccessibilityNodeInfo;"
        },
        {
            "class": "Landroid/view/accessibility/AccessibilityNodeInfo;",
[...]

PR #247, opened 2021-09-10
Detect APIs in the extended classes

haeter525:
**Description**
Refer to #158 .
This PR implements [the conclusion](https://github.com/quark-engine/quark-engine/issues/158#issuecomment-826026814) discussed in that issue.

> 1. If the rule uses the Android native API, we first check whether other classes inherit from this class, and then check whether the corresponding method is called.
> 2. To confirm the inherited method does not modify its original functionality, an extra checking is to see if the inherited method got overridden or not.

**Code changes**
1. baseapkinfo.py, apkinfo.py, and rzapkinfo.py
   - Rename method class_heriarchy
   - Add a method to get the extended classes of each class.
2. quark.py
   - Add a method to search APIs in the extended classes.

**Test plans**
- [ ] All tests passed.


pulorsok:
LGTM!
Nice work @haeter525, and thank you for the PR.

Issue #129, opened 2021-02-17
Make Quark read JSON rules faster


krnick:
Currently, quark reads all JSON rules in the directory one by one, but the number of rules gradually increases, which will cause a large increase in file IO and reduce speed. We can compress all rules into a single file or other forms of files to quickly read the rules.

PR #248, opened 2021-09-16
Reduce file IO and load rules in array

sharteeya:
**Description:**
- Solve #129 
- Prevent Quark from keeping reloading rule files in `cli.py`. Rules now only reload when necessary.
- Multiple rules can save into a single file now by using the `[]` format (JSON array). Of course, Quark can read those rules too.
- `RuleObject` can be created with loaded JSON data now.

**Code changed:**
- cli.py
- ruleobject.py


pulorsok:
@sharteeya LGTM, Thank you for the PR! 👍

PR #273, opened 2021-10-19
add save png funcionality of behaviors_comparison_radar_chart function

matteodalgrande:
In this fashion, it is possible to save a png when we use "Behaviors Comparison Radar Chart", that is really useful if someone decide to use quark-engine as docker.

### DIFF
--- a/quark/utils/graph.py
+++ b/quark/utils/graph.py
@@ -3,6 +3,7 @@
 # See the file 'LICENSE' for copying permission.
 
 import sys
+import os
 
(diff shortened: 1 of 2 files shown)


pulorsok:
@matteodalgrande LGTM, thanks for your contribution. 😄

krnick:
I don't think this PR should be merged since the package `kaleido` is not used in any codebase at all, which will cause problems when packaging in Debian. Why did you merge this without consideration?

Issue #311, opened 2022-03-05
Max num_process should have a check.


PaulNicolasHunter:
The minimum number of process is 1 and the max doesn't seem to have any limit, it should be multiprocessing.cpu_count(), machine may run out of memory [1] if exessive numbers are provided.

I suggest we put some sort of check on the [num processes](https://stackoverflow.com/questions/20039659/python-multiprocessings-pool-process-limit#:~:text=but%20an%20insane%20amount%20of%20processes%20started%20at%20once%20will%20cause%20death%20to%20the%20system%20because%20of%20the%20running%20out%20of%20memory).

Possible files - parrallelquark.py and cli.py

[1]: num processes

haeter525:
Hi, @PaulNicolasHunter.
> I suggest we put some sort of check on the [num processes](https://stackoverflow.com/questions/20039659/python-multiprocessings-pool-process-limit#:~:text=but%20an%20insane%20amount%20of%20processes%20started%20at%20once%20will%20cause%20death%20to%20the%20system%20because%20of%20the%20running%20out%20of%20memory).

Sure thing! Excessive processes over the available cores may exhaust the memory. Restricting it can avoid that situation.

Thank you for reporting this! If possible, would you like to implement the patch for it?

PaulNicolasHunter:
Sure @haeter525, will work on it.

haeter525:
Resolved by #312 .

PR #312, opened 2022-03-08
set num_of_process to a feasible value

PaulNicolasHunter:
w.r.t issue [#311](https://github.com/quark-engine/quark-engine/issues/311), number of process has been set to a value which is less or equal to the cpu_count - 1.


quark/core/parallelquark.py
```diff
@@ -113,7 +115,8 @@ def _apply_analysis_result(self, rule_obj):
     def __init__(self, apk, core_library, num_of_process=1):
         self._result_map = {}
         self._pool = Pool(
-            num_of_process - 1, self._worker_initializer, (apk, core_library)
+            min(num_of_process, cpu_count - 1), self._worker_initializer,
```
haeter525, line 118: `cpu_count` is a function. You should call it to get the number of cores.

haeter525 (changes requested):
Hi @PaulNicolasHunter.
I have left a comment on this change. Please test the code first before submitting a PR.

quark/core/parallelquark.py
PaulNicolasHunter: thanks, silly mistake.

haeter525 (approved):

haeter525:
@PaulNicolasHunter, LGTM! Nice work.
Since we mainly develop on branch master, I will change the target branch and solve the conflicts. Thanks!

PR #326, opened 2022-04-19
Add Radiocontrast feature

pulorsok:
**Description:**
Add the "Radiocontrast" feature that allows Quark to generate rules with the native APIs in a specific method.

**Usage:**
Here is the simplest way for using Radiocontrast API.
```python
from quark.radiocontrast import RadioContrast

# The target APK.
APK_PATH = "~/apk-malware-sample/Ahmyth.apk"

# The method that you want to generate rules. 
TARGET_METHOD = "Lahmyth/mine/king/ahmyth/CameraManager;->startUp(I)"

# The output directory for generated rules.
GENERATED_RULE_DIR = "~/generated_rules"

radiocontrast = RadioContrast(
    APK_PATH, 
    TARGET_METHOD, 
    GENERATED_RULE_DIR
)
radiocontrast.rule_generate()
```

**File changed:**
- Add `radiocontrast.py` into `quark-engine/quark/`.

### DIFF
--- a/quark/radiocontrast.py
+++ b/quark/radiocontrast.py
@@ -1,0 +1,142 @@
+# This file is part of Quark-Engine - https://github.com/quark-engine/quark-engine
+# See the file 'LICENSE' for copying permission.
+
+import os
+import json
+
+from tqdm import tqdm
+from quark.core.quark import Quark
+from quark.core.struct.ruleobject import RuleObject
+
+
+class RadioContrast:
+    """
+    This module is for generating rules with the APIs in a specific method.
+    """
+
+    def __init__(self, apk_path, target_method, output_dir, max_search_layer=3):
+        self.quark = Quark(apk_path)
+        self.apkinfo = self.quark.apkinfo
+
+        # Parse smali code into classname methodname and descriptor.
+        classname = target_method.split("->")[0]
+        methodname = target_method.split("->")[1].split("(")[0]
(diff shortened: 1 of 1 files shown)


quark/radiocontrast.py
```diff
@@ -0,0 +1,138 @@
...
+            class_name=classname, method_name=methodname, descriptor=descriptor,
+        )
+
+        if self.method is None:
+            print("ERROR: start method or end method not found")
+            return
+
+        self.output_dir = output_dir
+        self.api_set = set()
+        self.MAX_SEARCH_LAYER = 3
```
haeter525, line 36: I suggest adding a default argument to the method `__init__`. In this way, users can change the value of the variable MAX_SEARCH_LAYER if they want.

haeter525 (changes requested):

quark/radiocontrast.py
```diff
@@ -0,0 +1,137 @@
...
+                    tqdm.write(
+                        "{} and {} combination has some error when analyzing,\
+                        ERROR MESSAGE: {}".format(
+                            api1, api2, e,
+                        ),
+                    )
+                    continue
+
+                if comb.check_item[4]:
+                    rule_name = f"{outter_loop.n}{num}.json"
```
haeter525, line 125: It seems like the filenames of the generated rules are random numbers, could you name them sequentially? For example, `0.json`, `1.json`, `2.json`, and so on.

PR #337, opened 2022-04-25
Ensure Quark runs with a compatible Rizin

### LINKED ISSUE #332: Ensure Quark runs with the compatible Rizin
**Is your feature request related to a problem? Please describe.**

According to [the policy of release versioning](https://github.com/rizinorg/rizin/discussions/757#discussion-3247581), the current Rizin API (0.x.y) should not be considered stable since it's under initial development.

> Major version zero (0.y.z) is for initial development. Anything MAY  change at any time. The public API SHOULD NOT be considered stable.

Thus, we need to ensure Quark uses the compatible Rizin version to perform the analysis.

**Describe the solution you'd like**

The solution involves two parts - 

1. Check the Rizin installed in the user's environment. We inform the user whether the Rizin instance is compatible with Quark.
2. Provide a script to install Rizin easily. This way, users can have a Rizin installed independently from their system. Also, Quark can manage it regardless of the environment.

haeter525:
**Description**
Fix #332.
This PR aims to -
+ determine which Rizin executable Quark should use for the Rizin-based analysis.
+ compile and install an independent Rizin if necessary.

It consists of three steps to find a Rizin executable.
1. Check if the user specifies a path for Rizin (by option --rizin-path).
2. Check if a compatible Rizin executable exists in the system PATH.
3. Otherwise, install an independent Rizin in the Quark directory (~/.quark-engine) and use it for the analysis.

Note that a user can use the flag `--disable-rizin-installation` to disable the installation of the independent Rizin.
In this way, Quark will terminate the execution if no compatible Rizin exists.

**Test Plans**
- [x] Add seventeen tests for this logic.
- [ ] Ensure all tests pass.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -31,8 +31,8 @@ Once you see the following msg, then you're all set::
       -o, --output FILE               Output report in JSON
       -a, --apk FILE                  APK file  [required]
       -r, --rule PATH                 Rules directory  [default:
-                                      /Users/$USER/.quark-engine/quark-rules]
-      -g, --graph                     Create call graph to call_graph_image
+                                      /home/haeter/.quark-engine/quark-rules]
+      -g, --graph [png|json]          Create call graph to call_graph_image
                                       directory
       -c, --classification            Show rules classification
       -t, --threshold [100|80|60|40|20]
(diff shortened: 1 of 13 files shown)


pulorsok:
Hi @haeter525,
I think for better user experience.
We should simplify the steps for this feature.

I suggest we can adjust it to:
1. Check if the user environment installed compatible Rizin.
2. If not, ask the user, 
"do you want to install Rizin version x.x?" or 
"do you want to specify Rizin path?".
And let the user input answer.

In that case, we don't need to add new click options `--rizin-path` and `--disable-rizin-installation`. 
I think it is more intuitive for the user.
What do you think?

haeter525:
Thanks for the suggestion, @pulorsok.
I agree that inputting the answers is more intuitive than choosing the flags. I will update the PR.

PR #345, opened 2022-05-18
Add new feature for generate Quark web report

pulorsok:
Add new feature for generate Quark report.
With the following command, we can easily analyze the Android sample and output the web report.

See the demo [here](https://pulorsok.github.io/ruleviewer/web-report-demo).

```bash
quark -a sample.apk -s -w quark_report.html
```


quark/cli.py
```diff
@@ -372,6 +380,24 @@ def entry_point(
         with open(output, "w") as file:
             json.dump(json_report, file, indent=4)
             file.close()
+        
+    # Generate web report
+    if webreport:
+        if isinstance(data, ParallelQuark):
+            data.apply_rules(rule_buffer_list)
```
haeter525, line 387: I suggest removing the method call because we don't need it to generate reports. The method informs the `ParallelQuark` module to prepare for analyzing the given rule.

quark/webreport/generate.py
```diff
@@ -0,0 +1,282 @@
...
+        self.json_report = json_report
+
+        # Load html layout
+        with open("quark/webreport/upper.html") as file:
+            self.upperHTML = file.read()
+            file.close()
+
+        with open("quark/webreport/lower.html") as file:
+            self.lowerHTML = file.read()
+            file.close()
```
haeter525, line 17: The PyPi package by default doesn't include non-python files ([ref](https://setuptools.pypa.io/en/latest/userguide/datafiles.html?highlight=INclude#data-files-support)), so we may need to
+ modify the `package_data` keyword in [setup.py](https://github.com/quark-engine/quark-engine/blob/master/setup.py) to include these HTML
+ access these files by using [the recommended API](https://setuptools.pypa.io/en/latest/userguide/datafiles.html?highlight=INclude#accessing-data-files-at-runtime)

Please refer [here](https://github.com/quark-engine/quark-engine/pull/296/files) for an example.

haeter525 (changes requested):

haeter525 (approved):
LGTM! Nice feature!

Issue #324, opened 2022-03-30
Quark script implementation


18z:
Hi @PaulNicolasHunter 

@pulorsok , @haeter525 and I discuss the idea of #313 .
And we found out that the idea has high complexity and low practicality.

Therefore, we have a simple, concrete and practical idea for your GSoC 2022 proposal 
which is to implement ```Quark Script```.

The idea emerged when we used Quark to actually analyze Android applications.
It is often that we are still confused when we get Quark's summary reports.

For example, we'd like to know whether particular behavior appears 
in the third-party module that this application imported. 
Or whether this behavior performs right exactly in the application the author coded. 

Sometimes, we also want to know the relationships between the ```behaviors```.
For example, does X behavior trigger the Y behavior. 
Or is X triggered by Z?

And many other information behind the behavior 
we'd like to know during the application analysis.

So, the new idea in short, is to develop functions 
retrieving the information behind the behavior. 

Once the functions are implemented, we then can write some sort of ```Quark Scripts``` 
to quickly narrow down and explore the behaviors Quark detected.

For example:

```python
if isFound("001.json") and isNotThirdParty("001.json"):
    print("001.json is the behavior we're looking for.")
``` 

This script helps us to quickly verified 
if the behavior detected is not from the third party module.

```python
if isFound("001.json") and isFound("002.json"):
    if hasMutualParentFunction("001.json", "002.json"):
        print("001.json and 002.json performs under the same function.")
```

```python
if hasStrings("001.json", "hello world"):
    print("001.json contains strings hello world.")
```

So, what do you think @PaulNicolasHunter ? 😄

haeter525:
Moreover, implementing Quark Script is not complex work, but it brings more potential to the GSoC proposal.

For example, apply this feature to identify possible malware, which could be a notable contribution to the GSoC event. 💪

PaulNicolasHunter:
@18z @pulorsok @haeter525 Sounds like an Amazing idea guys! It makes sense to identify behaviour of the rules and their relation, seems like it will make the report more concrete and will also remove any possibilities of false positives. 
Looking forward to work on this, how do I get started? 😃

PaulNicolasHunter:
got it, will work on them

haeter525:
@PaulNicolasHunter. Can't wait to see this come true! 😁

Here is my proposal for GSoC 2021. You can follow those questions and finish yours.
[...]

PR #381, opened 2022-09-08
Add Quark Script APIs to detect CWE-749

haeter525:
#### Description
Please refer to #324.

This PR adds the following Quark script API to detect exposed dangerous methods or functions([CWE-749](https://cwe.mitre.org/data/definitions/749.html)).

1. `methodInstance.getArguments(none)`

**Test Plans**

- [x] All tests passed


pulorsok:
I suggest we can add the API `isArgumentTrue` to methodInstance rather than using `isTrue`.
The `isArgumentTrue` will first get the parameter values from the target method, then check if the value is True. I think it is more intuitive for the user.

What do you think?

PR #421, opened 2022-11-08
Update docs about release process in organization.rst

### LINKED ISSUE #417: Add docs about quark version release process
**Why**

Keep every release processed through the same workflow.

**How**

* Add docs about the quark version release process to Developments docs.

sidra-asa:
Close #417.

This PR updates the release process in docs with much specific notes.
Here's the new process.

Screen Shot 2022-11-08 at 1 44 12 PM


haeter525:
Hi @sidra-asa . Thanks for enhancing the process. However, I think there are some things improvable.

1. I suggest changing the description of the first step to "create an issue" and describing we use a checklist in the note.

2. Some notes only duplicate the steps. Please adjust them.

3. Please use the term Quark Script, not CWE script.

4. Since MobSF has its process to update the Quark version, we don't need to check the integrated tool in MobSF.

5. Please align all the notes. And please ensure the graph uses the same format as the one describing the Triage Process.

What do you think?

Issue #437, opened 2022-12-22
Dependency Kaleido installation failed in the build test of macos in the smoke test 


zinwang:
Dependency Kaleido installation failed in the build test of macos 12 with python 3.9 in the smoke test

PR #438, opened 2022-12-22
Remove dependency kaleido temporarily

zinwang:
This PR is related to Issue #437.
We remove dependency kaleido temporarily to avoid failure in build test.


haeter525 (approved):
LGTM!

PR #489, opened 2023-03-14
New Quark Method Explanation for method_recursive_search

PoJenC:
# method_recursive_search

**The algorithm of find_intersection**

This code searches for the intersection of method usage between two sets and recursively searches for upper-level method usage until it finds the same usage or reaches the maximum search depth.

Here is the process of `method_recursive_search`.

``` text
1. The method named method_recursive_search accepts three parameters depth, first_method_set, and second_method_set.
2. If the current recursive layer exceeds the MAX_SEARCH_LAYER, it returns None.
3. The first method set and second method set are copied to the next level, and next_level_set_1 and next_level_set_2 variables are created respectively.
4. The xrefs of methods in the first method set and the second method set are extended to the next layer. If any of these methods have upper functions, they are added to next_level_set_1 and next_level_set_2.
5. The method recursively calls itself, passing the next level parameters next_level_set_1, next_level_set_2, and depth+1.
6. If it finds the same method usage in the next level, it returns their intersection.
7. If it does not find the same method usage in the next level, it continues to search in the next layer until the recursive depth exceeds the maximum limit.
```

**The code of method_recursive_search**

``` python
def method_recursive_search(
    self, depth, first_method_set, second_method_set
):
    # Not found same method usage, try to find the next layer.
    depth += 1
    if depth > MAX_SEARCH_LAYER:
        return None

    # Append first layer into next layer.
    next_level_set_1 = first_method_set.copy()
    next_level_set_2 = second_method_set.copy()

    # Extend the xref from function into next layer.
    for method in first_method_set:
        if self.apkinfo.upperfunc(method):
            next_level_set_1 = (
                self.apkinfo.upperfunc(method) | next_level_set_1
            )
    for method in second_method_set:
        if self.apkinfo.upperfunc(method):
            next_level_set_2 = (
                self.apkinfo.upperfunc(method) | next_level_set_2
            )

    return self.find_intersection(
        next_level_set_1, next_level_set_2, depth
    )
```


zinwang:
@PoJenC 
Here are some problems I've found.

**1. The meaning is not precise.**
For example, `... method usage between two sets` does not illustrate what kind of sets they are.
And I suggest you check the semantics by translating it into Chinese.

**2. The word choice is not precise.**
    For example, `method usage` is unclear.
    And I suggest you check whether there are better alternatives for each word.

**3. The terminology is inconsistent.**
    For example,   `next level` and `next layer` are the same.
And I suggest you use the same word for the same concept.
    
**4. The text is not concise.**
     For example,  `maximum search depth`  should be `maximum depth` .
     And I suggest you avoid  unnecessary words and sentences that are hard to read.

I recommend you check the whole text you submitted and make the problems above extinguished.

pulorsok (approved):

pulorsok:
LGTM

zinwang (approved):
LGTM!

Issue #490, opened 2023-03-15
The latest document failed to build in readthedocs


zinwang:
The latest document failed to build in readthedocs. Refer to [here](https://readthedocs.org/projects/quark-engine/builds/19787528/).

PR #491, opened 2023-03-15
Fix the failure of building documents in readthedocs

zinwang:
Refer to #490


haeter525 (approved):
@haeter525 approves this patch! Thank @zinwang!

PR #506, opened 2023-05-11
Fix smoke test error due to image macos 10 being deprecated.

sidra-asa:
**File changed**

- .github/workflows/smoke_test.yml


haeter525 (approved):
LGTM! @haeter525 approved. Thank @sidra-asa for the quick fix.

PR #511, opened 2023-05-26
Add method reference of check_parameter_values

Daisu27:
check_parameter_values
==========================

**The algorithm of check_parameter_values**

The function ``check_parameter_values`` is designed to check if the parameter values in the source string match the specified patterns and keywords. Then it collects the matched strings into a set and return it.

Here is the process of ``check_parameter_values``.

    1. Create an empty set matched_string_set.

    2. Use tools.get_parenthetic_contents to extract the content that matches each pattern in the pattern_list from the source_str. Store the results in the parameter_strs list.

    3. Use zip to pair up the parameter_strs and keyword_item_list and iterate over them.

    4. For each pairing of parameter_str and keyword_item, perform the following operations:
        - Check if keyword_item is not None.
            - For each keyword in keyword_item, perform the following operations:
                - Check If regex is True, 
                    - If True, 
                        - Use re.findall to search for matching strings and store them in matched_strings.
                        - Check if matched_strings has any matching strings.
                            - If True, Add all nonempty string from matched_strings to the matched_string_set. 
                    - If False, add all keyword in parameter_str to the matched_string_set.
        
    5.  Once the iteration finishes, return a list of strings from the matched_string_set, which represents all the matched results.

Here is the flowchart of ``check_parameter_values``.

**The code of check_parameter_values**

```python

    @staticmethod
    def check_parameter_values(
        source_str, pattern_list, keyword_item_list, regex=False
    ) -> List[str]:
        matched_string_set = set()

        parameter_strs = [
            tools.get_parenthetic_contents(
                source_str, source_str.index(pattern) + len(pattern)
            )
            for pattern in pattern_list
        ]

        for parameter_str, keyword_item in zip(
            parameter_strs, keyword_item_list
        ):
            if keyword_item is None:
                continue

            for keyword in keyword_item:
                if regex:
                    matched_strings = re.findall(keyword, parameter_str)
                    if any(matched_strings):
                        matched_strings = filter(bool, matched_strings)
                        matched_strings = list(matched_strings)

                        element = matched_strings[0]
                        if isinstance(
                            element, collections.abc.Sequence
                        ) and not isinstance(element, str):
                            for str_list in matched_strings:
                                matched_string_set.update(str_list)

                        else:
                            matched_string_set.update(matched_strings)

                else:
                    if str(keyword) in parameter_str:
                        matched_string_set.add(keyword)

        return [e for e in list(matched_string_set) if bool(e)]
```

### DIFF
--- a/docs/source/quark_method_reference.rst
+++ b/docs/source/quark_method_reference.rst
@@ -541,12 +541,7 @@ Here is the process of ``check_parameter_values``.
                     - If True, 
                         - Use re.findall to search for strings in parameter_str that match the keyword.
                         - Check if matched_strings has any matching strings.
-                            - If True, 
-                                - Use filter to remove empty strings from matched_strings.
-                                - Check if first element is a sequence type and not string type.
-                                    - If True, Iterate over each string_list in matched_strings.
-                                        - Add each string from the str_list to the matched_string_set. 
-                                    - if False, add all the elements from matched_strings to matched_string_set. 
+                            - If True, Add all nonempty string from matched_strings to the matched_string_set. 
                     - If False, add all keyword in parameter_str to the matched_string_set.
         
     5.  Once the iteration finishes, return a list of strings from the matched_string_set, which represents all the matched results.
--- a/docs/source/quark_method_reference.rst
+++ b/docs/source/quark_method_reference.rst
@@ -522,7 +522,7 @@ check_parameter_values
 
 **The algorithm of check_parameter_values**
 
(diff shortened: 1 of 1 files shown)


sidra-asa:
@Daisu27 

Thank you for this PR.
I have a few suggestions.

1. Could it be better using `The function check_parameter_values checks whether the parameter values from a given string match the specified patterns and keywords.` in first paragraph of description ?
2. It seems that the subjection is missing in the second clause of `Use re.findall to search for matching strings and store in matched_strings.`.
3. The Noun should be Plural Nouns like `all nonempty strings` as well as `all keywords`.

sidra-asa (approved):
LGTM

pulorsok (approved):

Issue #516, opened 2023-05-31
CI failed due to an incompatible numpy package version installed by setup.py


zinwang:
numpy 1.25 installed using setup.py relies on Python 3.9 or above, resulting in an error when testing Python 3.8 in CI.

haeter525:
Close as #517 is merged.

PR #517, opened 2023-05-31
Install packages with pip instead of Setuptools in CI

zinwang:
Refer to Issue #516.


haeter525 (approved):
LGTM! @haeter525 approves. Thank @zinwang for this patch.

PR #537, opened 2023-08-18
Fix image link error in document of Visual Quark Script

pulorsok:
as title


haeter525 (approved):
LGTM! @haeter525 approves this patch!

Issue #555, opened 2023-10-02
GitHub generates security alerts due to outdated dependencies


zinwang:
GitHub generates security alerts due to outdated versions of some dependent packages specified in the pipfile.

PR #556, opened 2023-10-02
Fix GitHub security alerts by escalating versions of dependent packages.

zinwang:
Refer to Issue #555


haeter525 (approved):
Thank @zinwang for updating the dependencies versions! @haeter525 approves.

Issue #450, opened 2023-01-05
Match_keywords in rule not working as expected


cryptax:
Rule `00189.json` specifies a _match keyword_ for **SMS** and **body**. The rule is apparently however **matching any occurrence** of `ContentResolver.query` and `.getColumnIndex` - **even those without the keywords**.

**To Reproduce**
1. Use Quark v22.12.1
2. Download malicious sample `87127748e919d78927afc8f7bbf7725b36499abb5ea934a6a6af4012ce341660`  (Get it from Pithus: https://beta.pithus.org/report/87127748e919d78927afc8f7bbf7725b36499abb5ea934a6a6af4012ce341660)
3. Do `quark -a ./smssend.apk -d /root/.quark-engine/quark-rules/rules/00189.json `
4. Notice rule 00189 matches 100% for `com/software/android/install/DeviceRegistrar; makeContactsJsonData`
5. Decompile the app, and have a look at the corresponding code.

```java
 private static String makeContactsJsonData(Context context) {
        Uri uri = ContactsContract.CommonDataKinds.Phone.CONTENT_URI;
        Cursor cursor0 = context.getContentResolver().query(uri, new String[]{"display_name", "data1"}, null, null, null);
        int contact_name = cursor0.getColumnIndex("display_name");
        int contact_phonenumber = cursor0.getColumnIndex("data1");
        JSONObject jObj = new JSONObject();
        ...
```
The 2 matching lines are 2nd and 3rd line of the method. Notice however they are **not retrieving SMS but Contact **information (contact name and contact phone number).

So, either the **matching keyword section of the rule is not working** as expected (error of Quark engine), or the rule itself is wrong (remove match keywords? rename the rule?). Looks like the first.

```
"match_keywords":[
                "sms"
            ]
```

**Expected behavior**
In that particular case, if the rule is kept as such as matching_keywords work as I think they do, the rule should **not** match for the sample.

**Desktop (please complete the following information):**
 - OS: Ubuntu 22.04 LTS
 - Version 22.12.1

zinwang:
Hi @cryptax 
Thank you for reporting this!
We will start investigating it.
And I will keep you updated if there is anything new.

zinwang:
Hello,  @cryptax!
We have merged PR #561 to fix this issue.
Please check again to see if the problem has been solved.

PR #561, opened 2023-10-16
Fix keyword matching in fifth stage of quark analysis

zinwang:
Refer to Issue #450.


haeter525 (approved):
LGTM! Thank @zinwang for this patch.

Issue #564, opened 2023-10-20
There are typos in the quark script for CWE-749 and CWE-780


zinwang:
In the quark script for CWE-749, the parentheses in the last line are mismatched.
 Also, In the quark script for CWE-780, `useCryptographicAlgo` is misspelled into `useCryptographicAlgor`.

PR #565, opened 2023-10-20
Fix typo in quark script for  CWE-749 and CWE-780

zinwang:
Refer to Issue #564.


haeter525 (approved):
Nice patch. Thank @zinwang for this work. @haeter525 approves.

Issue #578, opened 2023-11-22
The compilation of radare2 failed in CI


zinwang:
The CI failed to compile radare2 cloned from its GitHub repo.

PR #579, opened 2023-11-22
Change the source of radare2 in CI

zinwang:
Refer to Issue #578.


haeter525 (approved):
LGTM. Thank @zinwang for this quick patch.

PR #590, opened 2023-12-12
Add method reference for select_label_menu

zinwang:
##  select_label_menu

**The algorithm of select_label_menu**

The ``select_label_menu`` method allows users to select labels shown in the radar chart.

```
1. Show a checkbox dialog for label selection.

2. Check if the user cancels the selection.
    - If YES, print "Canceled!" and exit.
    - If No, continue with the following steps.  
    
3. Check the number of labels the user selects is within the specified range.
    - If YES, return the labels.
    - If No, go back to the first step
```

Here is the flowchart of ``select_label_menu``.

**The code of select_label_menu**

```python
def select_label_menu(all_labels, min_labels=5, max_labels=10):
    """
    allows user to select label to be shown in radar chart
    :param all_labels: all label found on the rules
    :param min_labels: min label to be shown on radar chart (default 5)
    :param max_labels: max label to be shown on radar chart (default 10)
    :return: label selected
    """

    value_pair = [(label, label) for label in all_labels]

    while True:
        results_array = checkboxlist_dialog(
            title="Label-base Report",
            text=f"Select number of labels between {min_labels} and {max_labels}",
            values=value_pair,
        ).run()

        if results_array:

            if min_labels <= len(results_array) <= max_labels:
                break
        else:
            # user selects "Cancel" to leave the program.
            print("Canceled!")
            sys.exit()

    return results_array
```

### DIFF
--- a/docs/source/quark_method_reference.rst
+++ b/docs/source/quark_method_reference.rst
@@ -1896,7 +1896,7 @@ The ``select_label_menu`` method allows users to select labels shown in the rada
 
 Here is the flowchart of ``select_label_menu``.
 
-.. image:: https://i.imgur.com/JQm8CxP.png
+.. image:: https://i.imgur.com/ww5drta.png
 
 
 **The code of select_label_menu**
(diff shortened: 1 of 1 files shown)


haeter525 (changes requested):
Hi @zinwang. The steps don't describe the code correctly. For example, step 3 should be step 2. Please recheck the steps. Thanks.

haeter525 (approved):
LGTM. @haeter525 approves!

pulorsok (approved):
LGTM

Issue #599, opened 2024-01-06
Quark cannot analyze pivaa.apk when using radare2 as the core library.


sidra-asa:
**Describe the bug**
I analyzed the [pivaa.apk](https://github.com/quark-engine/apk-samples/blob/master/vulnerable-samples/pivaa.apk) with radare2 as the core library. The analysis was failed and the message showed that Quark failed to parse the permissions.

**To Reproduce**

```Bash
$ quark -a pivaa.apk -s --core-library radare2        

    ________                      __
    \_____  \  __ _______ _______|  | __
     /  / \  \|  |  \__  \_  __ \  |/ /
    /   \_/.  \  |  // __ \|  | \/    <
    \_____\ \_/____/(____  /__|  |__|_ \
           \__>          \/           \/ v23.12.1
    
                An Obfuscation-Neglect Android Malware Scoring System
                
  0%|                                                   | 0/211 [00:00<?, ?it/s]
Traceback (most recent call last):
  File "/Users/sasaki/.local/share/virtualenvs/quark-engine-KzHyODU-/bin/quark", line 33, in <module>
    sys.exit(load_entry_point('quark-engine', 'console_scripts', 'quark')())
  File "/Users/sasaki/.local/share/virtualenvs/quark-engine-KzHyODU-/lib/python3.8/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "/Users/sasaki/.local/share/virtualenvs/quark-engine-KzHyODU-/lib/python3.8/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/Users/sasaki/.local/share/virtualenvs/quark-engine-KzHyODU-/lib/python3.8/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/sasaki/.local/share/virtualenvs/quark-engine-KzHyODU-/lib/python3.8/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/Users/sasaki/program/python/quark-engine/quark/cli.py", line 334, in entry_point
    data.run(rule_checker)
  File "/Users/sasaki/program/python/quark-engine/quark/core/quark.py", line 458, in run
    elif set(rule_obj.permission).issubset(set(self.apkinfo.permissions)):
  File "/Users/sasaki/.pyenv/versions/3.8.16/lib/python3.8/functools.py", line 967, in __get__
[...]

PR #602, opened 2024-01-20
Fix permission analysis in the radare2 core for APKs with UTF-8 encoding

zinwang:
Refer to Issue #599 .


haeter525 (approved):
LGTM! Thank @zinwang for this patch. @haeter525 approves.

Issue #665, opened 2024-08-01
Inaccurate Parameter Extraction in getParamValue API


pulorsok:
**Description:**
The function `get_arguments_from_argument_str` within the Quark Script API `getParamValue` is designed to filter out irrelevant strings and extract the actual parameter values from the parameter string. These strings may contain information about the source methods and classes of the parameter values.

However, the current accuracy of this function's extraction is suboptimal, leading to the inclusion of non-actual parameter values in the extracted content.

Despite recent improvements to this API (see PR#1), which significantly increased the accuracy of parameter extraction from smali code, there are still some edge cases where the results may not be entirely accurate. Further refinement is planned for the future.

PR #666, opened 2024-08-01
Improve getParamValue API for More Accurate Parameter Extraction from Smali Code

pulorsok:
**Description:**
This pull request enhances the getParamValue API, which is responsible for extracting method parameters from smali code provided as strings. The primary focus of this update is to refine the filtering mechanism, ensuring cleaner and more accurate extraction of parameters without including extraneous characters.

Refer to issue #665.


haeter525 (approved):
LGTM. Thank @pulorsok.

Issue #679, opened 2024-08-30
Update minimum Python version to 3.9


zinwang:
Since Python 3.8 will reach the [end-of-life](https://devguide.python.org/versions/#supported-versions) in October, we should update the minimum Python version Quark supports to 3.9.

PR #680, opened 2024-08-30
Update minimum python version to 3.9

zinwang:
Refer to Issue #679


haeter525 (changes requested):
Thank @zinwang for this PR. Some suggestions below.

1. Could you help fix the conflict in README?
2. Please update [the CI in quark-script](https://github.com/quark-engine/quark-script/blob/main/.github/workflows/testQuarkScript.yml) so that it doesn't fail if we merge the PR.

haeter525 (approved):
LGTM. Thank @zinwang for this update!

Issue #710, opened 2024-11-07
Invalid Syntax errors in Python 3.12


ajinabraham:
**Describe the bug**

```
/mobsf-ss-py3.12/lib/python3.12/site-packages/quark/utils/output.py:156: SyntaxWarning: invalid escape sequence '\l'
  descriptions = "\l".join(report_dict[parent]) + "\l"
/mobsf-ss-py3.12/lib/python3.12/site-packages/quark/utils/output.py:156: SyntaxWarning: invalid escape sequence '\l'
```

**To Reproduce**
Run quark analysis in MobSF with Python3.12

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

zinwang:
Hi, thank you for reporting this issue! We have confirmed it and will look into it.

zinwang:
@ajinabraham 
Hi, we've pushed a patch to fix the issue and released a new version, [v24.11.2](https://github.com/quark-engine/quark-engine/releases/tag/v24.11.2). Please give it a try.

I'll close this issue. If the problem persists or you have further questions, feel free to reopen this issue.

ajinabraham:
Thanks for the quick turnaround!

Issue #701, opened 2024-10-21
Library `pkg_resources` is deprecated and removed in Python 3.12


haeter525:
**Describe the bug**

The Setuptools document states that [`pkg_resources` is deprecated and replaced by `importlib_resources`](https://setuptools.pypa.io/en/latest/pkg_resources.html). Also, [Python 3.12 has removed it](https://docs.python.org/3/whatsnew/3.12.html) from the standard library.

Therefore, running Quark on Python 3.12+ encountered the following error.

```
Traceback (most recent call last):
  File "/Users/shengfeng/codespace/test/.venv/bin/quark", line 5, in <module>
    from quark.cli import entry_point
  File "/Users/shengfeng/codespace/test/.venv/lib/python3.13/site-packages/quark/cli.py", line 13, in <module>
    from quark.core.parallelquark import ParallelQuark
  File "/Users/shengfeng/codespace/test/.venv/lib/python3.13/site-packages/quark/core/parallelquark.py", line 9, in <module>
    from quark.core.quark import Quark
  File "/Users/shengfeng/codespace/test/.venv/lib/python3.13/site-packages/quark/core/quark.py", line 16, in <module>
    from quark.core.rzapkinfo import RizinImp
  File "/Users/shengfeng/codespace/test/.venv/lib/python3.13/site-packages/quark/core/rzapkinfo.py", line 17, in <module>
    from quark.core.axmlreader import AxmlReader
  File "/Users/shengfeng/codespace/test/.venv/lib/python3.13/site-packages/quark/core/axmlreader/__init__.py", line 9, in <module>
    import pkg_resources
ModuleNotFoundError: No module named 'pkg_resources'
```

**To Reproduce**
1. Use an environment on Python 3.12 or later
2. Install Quark with `pip install -U quark`
3. Run Quark with `quark -v`

**To Resolve**

To get Quark running on Python 3.12+, we need to migrate to `importlib_resources`. [Here](https://importlib-resources.readthedocs.io/en/latest/migration.html#migration-guide) is an example.

We currently use `pkg_resources` in 3 files.
+ [`quark/webreport/generate.py`](https://github.com/quark-engine/quark-engine/blob/7e1f42d05839186b2a0a6f59850eb9febd7ccaf5/quark/webreport/generate.py#L1)
[...]

PR #712, opened 2024-11-12
Replace deprecated package and change syntax to solve issues in python 3.12

zinwang:
This PR aims to solve issue #701 and #710 .

### DIFF
--- a/quark/core/axmlreader/__init__.py
+++ b/quark/core/axmlreader/__init__.py
@@ -91,7 +91,8 @@ class AxmlReader(object):
         axmlDefinitionPath = (
             importlib_resources.files(base_path) / "axml_definition"
         )
-        with importlib_resources.as_file(axmlDefinitionPath) as axmlDefinitionFile:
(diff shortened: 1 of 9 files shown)


quark/core/axmlreader/__init__.py
```diff
@@ -104,6 +108,7 @@ def __init__(self, file_path, core_library="rizin", structure_path=None):
             self._core = r2pipe.open(file_path)
 
         self._core.cmd(f"pfo {structure_path}")
+        file_manager.close()
```
haeter525, line 111: Could you use `ExitStack` with a `with` statement to make the code clear?

For example,
```python
with ExitStack() as stack:
    # The code goes here
```

[Reference](https://docs.python.org/3/library/contextlib.html#contextlib.ExitStack)

haeter525 (changes requested):

haeter525 (approved):
LGTM. Thank @zinwang!

PR #797, opened 2025-08-16
Add CI Failure Advisor 

### LINKED ISSUE #796: Add a CI advisor for root cause Analysis and solution suggestions for CI failures
CI failures are a frequent source of frustration. When they occur before a scheduled release, we are forced to identify and fix the issue within a limited timeframe.  Having an AI advisor that triggers immediately upon a CI failure—performing an initial root-cause analysis and suggesting potential solutions—would significantly reduce the burden and help keep releases on schedule.

zinwang:
Resolves #796

### DIFF
--- a/.github/workflows/ci-failure-advisor.yml
+++ b/.github/workflows/ci-failure-advisor.yml
@@ -249,7 +249,7 @@ jobs:
         run: |
           bodyFile="analysis.md"
           callerRunCommitSha="${{ github.event.workflow_run.head_sha }}"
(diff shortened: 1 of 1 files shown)


zinwang:
If the failed CI workflow is triggered by a pull request, the advisor will post the report as a comment under that PR like [this](https://github.com/zinwang/quark-engine/pull/158#issuecomment-3193818220). 

Otherwise, the advisor will create an issue containing the report like [this](https://github.com/zinwang/quark-engine/issues/159).

PR #802, opened 2025-08-20
Add Quark Script APIs to detect CWE-359

zinwang:
The following three APIs are added to detect [CWE-359](https://cwe.mitre.org/data/definitions/359.html):

### BaseApkinfo.providers(self)
+ Description: Get provider elements from the manifest file.
+ return: python list containing provider elements 

### getProviders(samplePath)
+ Description: Get provider elements from the manifest file of the target sample.
+ params:
    1. samplePath: the file path of target sample
+ return: python list containing provider elements 

### providerInstance.isExported(none)

+ Description: Check if the provider element set ``android:exported=true``.
+ params: none
+ return: True/False

### DIFF
--- a/tests/core/test_apkinfo.py
+++ b/tests/core/test_apkinfo.py
@@ -750,7 +750,7 @@ def apkinfoPivaa(request, SAMPLE_PATH_pivaa, dex_file):
 
     yield apkinfo
 
-class AnotherTestApkinfo:
+class TestAnotherApkinfo:
     @staticmethod
     def test_providers(apkinfoPivaa):
         match apkinfo.ret_type:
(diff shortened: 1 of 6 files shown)


tests/core/test_apkinfo.py
```diff
@@ -723,3 +723,61 @@ def test_get_wrapper_smali(self, apkinfo):
...
+    if fileType == "DEX":
+        fileToBeAnalyzed = dex_file_pivaa
+
+    apkinfo = apkinfoClass(fileToBeAnalyzed)
+
+    yield apkinfo
+
+class TestAnotherApkinfo:
+    @staticmethod
+    def test_providers(apkinfoPivaa):
```
haeter525, line 770: Could you define this unit test under the `TestApkinfo` class? I don’t see a need to introduce another class for it.

haeter525 (changes requested):

PR #823, opened 2025-10-30
Optimize the document of Quark Script CWE-926

pulorsok:
CWE-926 Detection Process Using Quark Script API
========================

image

Let's use the above APIs to show how the Quark script finds this vulnerability.

First, we use Quark API ``getActivities(samplePath)`` to get all activity data in the manifest. Then, we use ``activityInstance.hasIntentFilter()`` to check if the activities have ``intent-filter``. Also, we use ``activityInstance.isExported()`` to check if the activities set the attribute ``android:exported=true``. If both are **true**, then the APK exports the component for use by other applications. That may cause CWE-926 vulnerabilities.

Quark Script: CWE-926.py
========================

image

```python
	from quark.script import *

	SAMPLE_PATH = "dvba.apk"

	for activityInstance in getActivities(SAMPLE_PATH):

	    if activityInstance.hasIntentFilter() and activityInstance.isExported():
		print(f"CWE-926 is detected in the activity, {activityInstance}")
```

### DIFF
--- a/docs/source/quark_script.rst
+++ b/docs/source/quark_script.rst
@@ -972,7 +972,7 @@ We use the `dvba.apk <https://github.com/rewanthtammana/Damn-Vulnerable-Bank>`_ 
 .. image:: https://imgur.com/KoOt5ii.jpg
 
 CWE-926 Detection Process Using Quark Script API
-========================
+================================================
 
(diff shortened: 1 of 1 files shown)


haeter525 (approved):
LGTM. Thanks.

PR #828, opened 2025-11-10
Fix behavior not detected in Slocker malware family

### LINKED ISSUE #827: Existing behavior is not detected with 100% confidence
**Describe the bug**

A behavior is not detected with 100% confidence for an APK (SHA256: `570e2811e8c87f714eb3485c271ec03b9de699c6b7f67e858a24396ce5f7b69e`), although the decompiled code shows the behavior does exist. The summary report is attached below.

Image

The Quark rule used for detecting this behavior:

```json
{
    "crime": "Create a window and add view to it",
    "permission": [],
    "api": [
        {
            "descriptor": "(I I I I I)V",
            "class": "Landroid/view/WindowManager$LayoutParams;",
            "method": "<init>",
            "match_keywords" : [
                "2002"
            ]
        },
        {
            "descriptor": "(Landroid/view/View; Landroid/view/ViewGroup$LayoutParams;)V",
            "class": "Landroid/view/WindowManager",
            "method": "addView"
        }
    ],
    "score": 2.11,
    "label": [
    ]
}
```

From the decompiled code, the behavior appears in the `onCreate` method of the `com.lololo.LockService` class. The first API call occurs at line 67 and saves the result to `layoutParams`. Then the second API call at line 72 uses that variable as the second argument.

Image

**To Reproduce**
1. Download the APK and the Quark rule.
2. Run `quark -a <PATH_TO_THE_APK> -s <PATH_TO_THE_RULE>`
3. Check the summary report

**Expected behavior**

Since the behavior clearly exists in the decompiled code, Quark should detect it with 100% confidence.

haeter525:
#### Description
This PR fixes #827 with the following changes:
1. Add support for invoke-direct/range.
2. Fix PyEval treating multiple registers referencing the same object as different objects.
3. Fix PyEval not tracking internal state changes of an object.
4. Reduce memory usage by removing duplicated parameters stored in PyEval.

**Tests**

- [x] All unit tests passed
- [x] Verify that the relevant issue is fixed.
    截圖 2025-11-11 02 11 20

### DIFF
--- a/quark/evaluator/pyeval.py
+++ b/quark/evaluator/pyeval.py
@@ -184,27 +184,31 @@ class PyEval:
                 value_of_reg_list.append(var_obj.value)
 
         # Remove duplicate parameter values to save memory
-        for value_idx in range(len(value_of_reg_list)):
-            for check_idx in range(value_idx + 1, len(value_of_reg_list)):
-                if value_of_reg_list[value_idx] == value_of_reg_list[check_idx]:
-                    value_of_reg_list[check_idx] = f"<ref_{value_idx}>"
+        seen = {}
+        for idx, val in enumerate(value_of_reg_list):
+            if val in seen:
+                value_of_reg_list[idx] = f"<ref_{seen[val]}>"
+            else:
+                seen[val] = idx
 
         invoked_state = f"{executed_fuc}({','.join(value_of_reg_list)})"
 
         # insert the function and the parameter into called_by_func
         for reg in reg_list:
             index = int(reg[1:])
-            obj_stack = self.table_obj.get_obj_list(index)
-            if obj_stack:
-                # add the function name into each parameter table
-                var_obj = self.table_obj.pop(index)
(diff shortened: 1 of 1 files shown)


quark/evaluator/pyeval.py
```diff
@@ -177,6 +183,12 @@ def _invoke(self, instruction, look_up=False, skip_self=False):
                 var_obj = self.table_obj.pop(index)
                 value_of_reg_list.append(var_obj.value)
 
+        # Remove duplicate parameter values to save memory
```
sidra-asa: Replaced the nested loops with a dictionary-based approach to track seen values.

For example.

```Python
seen = {} 
for idx, val in enumerate(value_of_reg_list): 
    if val in seen: 
        value_of_reg_list[idx] = f"<ref_{seen[val]}>" 
    else: 
        seen[val] = idx
```

zinwang (approved):
LGTM! Thank you @haeter525

PR #835, opened 2025-11-27
Add freshquark command to installation instructions

gemesa:
For newcomers it might not be obvious how to get the rules. Previously, if they followed the install guide and tried to analyze an APK, they were presented with the following error:
`Path '/<path>/.quark-engine/quark-rules/rules' does not exist.`


zinwang (approved):
Hi, @gemesa. Thank you for noticing the problem!
I've reviewed the changes, and they look great. Merging now!

Issue #299, opened 2022-01-09
quark gets stuck during the scan


javixeneize:
Hi

I am running quark in my local machine (mac m1) against this apk --> https://apkpure.com/es/nodo-play/nodo.play using the command  quark -a nodoplay.apk -d

When it reaches 52% of the scan, it gets stuck until it crashes

[1]    24031 killed     quark -a nodoplay.apk -d
/opt/homebrew/Cellar/python@3.9/3.9.9/Frameworks/Python.framework/Versions/3.9/lib/python3.9/multiprocessing/resource_tracker.py:216: UserWarning: resource_tracker: There appear to be 1 leaked semaphore objects to clean up at shutdown
  warnings.warn('resource_tracker: There appear to be %d '

Is this a known issue?

Thanks

haeter525:
Hi @javixeneize 

Thank you for reporting this issue!
We will start to investigate it soon. If there is anything new, I will keep you updated :)

javixeneize:
Thanks!

tirekin:
Hi,

I have the same issue with apk sha256: 

> 8bfa2948ff8f84b2620c672ccfec8efa28f738d849276f645429c11ec73a998d

The size of the APK is 2.3 MB.

Additionally, removing `00030.json`, `00142.json`, `00130.json` and `00129.json` did not solve the issue. There may be more rules which cause this behavior.

Command: `quark -a 8bfa2948ff8f84b2620c672ccfec8efa28f738d849276f645429c11ec73a998d.apk -s -t 80`

Quark version: `22.2.1`

zinwang:
The samples below also have this issue.


|index|sha256|
|--|--|
|1|022a01566d6033f6d90ab182c4e69f80a3851565aaaa386c8fa1a9435cb55c91|
|2|059b5f74e053c2966775157cd521580fcfaa3b1a7613560b8f499dbd9c11d4b4|
|3|0713a683567125ea6fdff233cfa850b36a0d2c7d7c964510405cbdf669fe2a8b|
|4|2c25a7fefac166cd897b5da58186dfe0b7d02df2fe28f1a07ccdd771a86f10b5|
|5|4b2b411e03aafaa19ea93286fadd39a5134f4a039db2d5019b1054547c0d5601|
|6|5c01f7727c78dea9c89dccf92b01b4c45e69406e6462340779401497bf4d4589|
[...]

PR #845, opened 2025-12-21
Record register usage instead of full state to reduce memory usage

haeter525:
This PR addresses #299 by improving memory efficiency during Quark's tainted analysis.

### Changes
1. **Taint Analysis Refinement**: Track register usage instead of full state snapshots.  
   - Files: `valuenode.py`, `registerobject.py`, `pyeval.py`  
1. **Quark Module & Quark Script API Updates**: Update code to reflect the above change.
   - Files: `quark.py`, `quark/script/__init__.py`  
3. **Crash Fixes**:
   - Remove register count limits to support methods with more than 40 registers (`tableobject.py`).  
   - Treat unknown-type registers as object-bearing to improve tracking accuracy (`registerobject.py`).  
   - Add keyword properties to `RuleObject` to simplify code (`ruleobject.py`).  

### Testing
- [x] Add unit tests for all changes.  
- [x] Ensure all unit tests pass.  
- [x] Confirm summary reports for APKs in [apk-samples](https://github.com/ev-flow/apk-samples) are either unchanged or improved in accuracy
- [ ] Confirm risk levels in [malware family reports](https://quark-engine.readthedocs.io/en/latest/quark_rules.html) remain "High Risk".  
  - ⚠️ **Two DawDropper samples downgraded to "Moderate Risk"** due to analysis result changes.
    1. `022a01566d6033f6d90ab182c4e69f80a3851565aaaa386c8fa1a9435cb55c91`  
    2. `02499a198a8be5e203b7929287115cc84d286fc6afdb1bc84f902e433a7961e4`
    A follow-up PR will address this issue.

#### Memory Usage
For APKs mentioned in #299, Quark can now complete analysis using at most **1 GB** of memory.  

| SHA256 | Peak Memory Usage(MB) |
|---|---:|
| 8ab9cbc73c6011244e651cff4d011c122bd29083e92cfbb7cd4e397cb6b22092 | 245.75 |
| 5c01f7727c78dea9c89dccf92b01b4c45e69406e6462340779401497bf4d4589 | 210.50 |
| 059b5f74e053c2966775157cd521580fcfaa3b1a7613560b8f499dbd9c11d4b4 | 210.51 |
| 8c365bd58edeb2ca371ead5e28350ee6c480a79f558d967ecbef525e9f1d7b3e | 211.50 |
| a1298cc00605c79679f72b22d5c9c8e5c8557218458d6a6bd152b2c2514810eb | 455.88 |
| eec5096dfca6824317863f9225c29f6c4b3442c48fefa62dc382e3569bca5a60 | 172.32 |
| dee1eaaa8879a7d321ef4e698203be7b23eeda80a6dea3c70cbf3138597b1800 | 210.47 |
| 0713a683567125ea6fdff233cfa850b36a0d2c7d7c964510405cbdf669fe2a8b | 786.75 |
| 4b2b411e03aafaa19ea93286fadd39a5134f4a039db2d5019b1054547c0d5601 | 793.43 |
| 8d1cf0eff5e73d4c5b9fda82268c436dae8dcd2166a86f76374afd537ad2962f | 998.77 |
| 2c25a7fefac166cd897b5da58186dfe0b7d02df2fe28f1a07ccdd771a86f10b5 | 1018.40 |
| f46b863952599b91a4d2d682a80f345dfa03fad473d1938f2c53a3139c87a019 | 792.21 |


sidra-asa:
@haeter525 

Thank you for the quick and high-quality refactoring. The current implementation looks solid.

One final suggestion for robustness: Could it be better to add a loop detection mechanism in iterativeResolve? 
While the cache helps , explicit detection would ensure the engine gracefully handles malicious APKs with cyclic structures.

Just a thought to make this even more bulletproof.

haeter525:
Hi @sidra-asa, good point. I’ve added loop detection to iterativeResolve. Thanks!

quark/evaluator/pyeval.py
```diff
@@ -148,21 +150,25 @@ def __init__(self, apkinfo):
...
+        Function call in Android smali code. It will check if the corresponding
+        table field has a value, if it does,
         inserts its own function name into called_by_func column.
         """
 
-        if look_up:
+        opcode, *regList, targetMethod = instruction
+        regIdxList = [int(r[1:]) for r in regList]
+        
+        if look_up and regIdxList and (firstReg := regIdxList[0]):
```
pulorsok: There is a concern here.

Based on the code, the purpose of this if condition appears to be resolving the actual method implementation for a given smali invoke instruction. However, I am not sure what the intended purpose of `(firstReg := regIdxList[0])` is in this context.

Assuming that `(firstReg := regIdxList[0])` is necessary, this style of condition introduces a potential issue:

If the smali instruction is `invoke-virtual {v0, v1, v2}`, then `regIdxList` will be `[0, 1, 2]`, and `firstReg` will be `0`. However, in Python, `0` is evaluated as `False`.

Therefore, even if `look_up` is `True`, when the first register is `v0` (index 0), the entire condition becomes `False`, causing the expected `_lookup_implement` call to be skipped, which in turn affects the method implementation resolution result.

haeter525: Hi @pulorsok, I've updated the precondition logic to address the issue. Thanks!

PR #866, opened 2026-02-25
Add safe DEX checksum repair flow (for androguard)

e619003:
### Description

When loading APK/DEX with the Androguard core, corrupted DEX header checksum/signature causes Androguard to refuse loading and abort the analysis.

### Workflow

Scope: This change is Androguard‑only. Other cores (rizin, radare2, shuriken) are unaffected.

1. Androguard load flow (APK path)
    - If AnalyzeAPK(self.data, raw=True) raises a checksum‑related error:
        1. When `--auto-fix-checksum` is provided:
            Silently repair all .dex headers in‑memory and retry loading.
            - DEX header repair:
                - SHA‑1 at 0x0C..0x1F = sha1(data[0x20:]) (20 bytes).
                - Adler32 at 0x08..0x0B = adler32(data[0x0C:]) (4 bytes, little‑endian).

        2. When `--auto-fix-checksum` is not provided:
            - **Interactive TTY** : prompt the user once with a clear warning that repairing modifies DEX headers and invalidates APK signatures (hashes will change). Proceed only if the user agrees.
            - **Non‑interactive** : skip repair and raise the original error, printing a concise hint to rerun with --auto-fix-checksum.

2. Androguard load flow (DEX path)
    - Same policy as the APK path:
        - If a checksum‑related error occurs in addDEX(...), honor --auto-fix-checksum to repair the single DEX header and retry.
        - Without the --auto-fix-checksum flag and prompt in TTY, skip in non‑interactive.

### Behavior

- Without `--auto-fix-checksum` (default)
    - **Interactive TTY** : shows a single confirmation prompt, repairs only on user consent.
    - **Non‑interactive** : does not repair and does not block; prints a hint to re‑run with --auto-fix-checksum, then raises the original error.

- With `--auto-fix-checksum`
    - Performs an in‑memory silent repair and retries the load (no prompt).
  
- Other cores (not androguard):
    - unchanged.

### DIFF
--- a/quark/cli.py
+++ b/quark/cli.py
@@ -151,7 +151,8 @@ logo()
 )
 @click.option(
     "--auto-fix-checksum",
-    help="Automatically repair damaged DEX checksum/signature before analyzing (androguard only).",
+    help="Automatically repair damaged DEX checksum/signature before analyzing (androguard only)." + 
+    "When not provided, Quark will prompt in interactive TTY and skip in non-interactive runs.",
     is_flag=True,
     default=False,
     show_default=True,
--- a/quark/core/apkinfo.py
+++ b/quark/core/apkinfo.py
@@ -2,11 +2,13 @@
 # This file is part of Quark-Engine - https://github.com/quark-engine/quark-engine
 # See the file 'LICENSE' for copying permission.
 
+import click
 import functools
 import hashlib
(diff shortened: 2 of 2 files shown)


pulorsok (approved):
LGTM

PR #872, opened 2026-02-27
Fix #871: Fix instruction names for PyEval.CAST_TYPE

### LINKED ISSUE #871: Incorrect namings of CAST_TYPE instructions in PyEval.eval
In the code, these instructions are named like "{type1}-{type2}", https://github.com/ev-flow/quark-engine/blob/b821a634c551c654ef271b91080fd5d73b2a9446/quark/evaluator/pyeval.py#L105

however, it should be "{type1}-to-{type2}", according to https://source.android.com/docs/core/runtime/dalvik-bytecode

Androguard and shuriken both have "{type1}-to-{type2}" format, so as to now quark skips those instructions

dolamroth:
Fix https://github.com/ev-flow/quark-engine/issues/871

### DIFF
--- a/.github/workflows/smoke_test.yml
+++ b/.github/workflows/smoke_test.yml
@@ -99,9 +99,11 @@ jobs:
 
     - name: Check Ahmyth Result
       shell: bash
-      # This sample should have 39 behaviors with 100% confidence
+      # This sample should have 39 behaviors with 100% confidence.
+      # Rule 00015 is currently detected with 80% confidence because PyEval fails
+      # to track byte arrays (issue #889: https://github.com/ev-flow/quark-engine/issues/889).
       run: |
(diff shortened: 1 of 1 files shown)


haeter525:
Hi @dolamroth. Your changes look great!

The test failed for two reasons, but <ins>**neither was caused by your changes**</ins>:

*   One was a **test baseline issue**. We recently introduced new rules to the built-in ruleset but didn't update the test to reflect that. (Now fixed in [#888](https://github.com/ev-flow/quark-engine/pull/888).)
*   The other is a **pre-existing bug** that your fix helped uncover. I've opened [#889](https://github.com/ev-flow/quark-engine/issues/889) to track it.

Your change is good to go. I'll apply a fix to your branch to get the tests to pass, then merge the PR.
Thank you for the patch! :D

haeter525 (approved):

PR #885, opened 2026-03-13
Fix annotation of quark.utils.tools.contains

dolamroth:
Fixes https://github.com/ev-flow/quark-engine/issues/880


haeter525 (approved):
LGTM. Thank you!

PR #933, opened 2026-06-24
ci: drop Shuriken-Analyzer build from CI

pulorsok:
## Why

The CI env-prep step installed Shuriken-Analyzer via:

```
pip install git+https://github.com/Fare9/Shuriken-Analyzer.git@main#subdirectory=shuriken/bindings/Python/
```

That compiles C++/pybind11 bindings at install time. The upstream Shuriken-Analyzer is unmaintained and the build has started **failing during CI setup**, blocking otherwise-unrelated PRs.

Shuriken is **not** a required (or extras) dependency of Quark — the CLI defaults to the `androguard` core library (`quark/cli.py`), and Shuriken stays an optional, user-installable core (`from shuriken import ...` is already guarded by `try/except ModuleNotFoundError` in `quark/core/shurikenapkinfo.py`).

## What

- `.github/workflows/pytest.yml`: drop the `Install Shuriken-Analyzer` step
- `.github/workflows/smoke_test.yml`: drop the Linux + macOS `Install Shuriken-Analyzer` steps
- `tests/core/test_apkinfo.py`: since pytest no longer installs Shuriken, guard the `ShurikenImp` parametrized fixtures (`apkinfo`, `apkinfoPivaa`) with `pytest.skip` when the `shuriken` module is absent — `ShurikenImp.__init__` otherwise raises, which would turn those cases into pytest ERRORs.

After the change, `grep -rni shuriken .github/` returns nothing; no other workflow or Dockerfile rebuilds it. Shuriken CLI support and its docs (`--core-library shuriken`) are intentionally left intact.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -60,10 +60,6 @@ jobs:
         sh install.sh -y
         cargo install project_ares@0.10.0
 
-    - name: Install Shuriken-Analyzer
-      run: |
-        pip install git+https://github.com/Fare9/Shuriken-Analyzer.git@main#subdirectory=shuriken/bindings/Python/
-
     - name: Install Quark-Engine
       run: pip install .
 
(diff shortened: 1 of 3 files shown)


haeter525 (approved):
LGTM.

PR #937, opened 2026-06-27
fix: validate default rule path only when used

### LINKED ISSUE #936: `--rule` default path validated even when unused
## Summary

## Reproduce
4. See error.
Image

## Expected behavior

## Root cause
```python
@click.option(
    "-r",
    "--rule",
    help="Rules directory",
    default=f"{config.DIR_PATH}",
    required=False,
    show_default=True,
)
```

## Fix

gaoflow:
## Summary
- stop Click from validating the default --rule path during option parsing
- validate the rule path only when Quark actually loads rules from that path
- add CLI regression tests for custom single-rule use and missing rule-directory use

Fixes #936.

## Tests
- uv run --with pytest --with requests --with click --with numpy --with tqdm --with prettytable --with colorama --with pathvalidate --with graphviz --with plotly --with-editable . pytest tests/test_cli.py -q
- uv run --with black black --check tests/test_cli.py
- git diff --check

Note: I did not run  because the current file has pre-existing formatting that newer Black would reflow outside this fix. I kept the production diff to the rule option and the runtime path check.

### DIFF
--- a/quark/cli.py
+++ b/quark/cli.py
@@ -66,7 +66,7 @@ logo()
     "-r",
     "--rule",
     help="Rules directory",
-    type=click.Path(exists=True, file_okay=True, dir_okay=True),
+    type=click.Path(exists=False, file_okay=True, dir_okay=True),
     default=f"{config.DIR_PATH}",
     required=False,
     show_default=True,
@@ -207,7 +207,7 @@ def entry_point(
     else:
         if not os.path.exists(rule):
             raise click.BadParameter(
-                f"Path {rule!r} does not exist.",
+                f"Path {rule} does not exist.",
                 param_hint="--rule",
             )
 
@@ -467,13 +467,6 @@ def entry_point(
 
     if isinstance(data, ParallelQuark):
         data.close()
-
-
-# Click validates option defaults before the command body can choose a rule source.
-for param in entry_point.params:
-    if param.name == "rule":
-        param.type = click.Path(exists=False, file_okay=True, dir_okay=True)
-        break
 
 
 def update_rule_buffer(rule_buffer_list, rule_path_list):
--- a/quark/cli.py
+++ b/quark/cli.py
@@ -205,6 +205,12 @@ def entry_point(
 
         rule_path_list = [rule_filter]
     else:
+        if not os.path.exists(rule):
+            raise click.BadParameter(
+                f"Path {rule!r} does not exist.",
+                param_hint="--rule",
+            )
+
         rule_path_list = [
             os.path.join(dir_path, file)
             for dir_path, _, file_list in os.walk(rule)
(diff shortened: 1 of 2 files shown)


quark/cli.py
```diff
@@ -463,6 +469,13 @@ def entry_point(
         data.close()
 
 
+# Click validates option defaults before the command body can choose a rule source.
+for param in entry_point.params:
+    if param.name == "rule":
+        param.type = click.Path(exists=False, file_okay=True, dir_okay=True)
+        break
```
haeter525: These lines patch the `--rule` option after the command is defined. Since the goal is to skip path validation, would it be simpler to set `exists=False` directly at line 69 instead?

https://github.com/ev-flow/quark-engine/blob/28b0ef7332bc92df953f336d87b99441a2b53626/quark/cli.py#L69

tests/test_cli.py
```diff
@@ -0,0 +1,100 @@
...
+    apk.write_text("")
+    monkeypatch.setattr(_rule_option(), "default", str(missing_default_rules))
+
+    runner = CliRunner()
+
+    with patch("quark.cli.Quark") as mock_quark:
+        data = mock_quark.return_value
+        data.quark_analysis.score_sum = 0
+        data.quark_analysis.weight_sum = 0
+        data.quark_analysis.summary_report_table = ""
```
haeter525: These lines duplicate the setup block at lines 91–95. Could you extract it into a `mock_quark` pytest fixture to deduplicate it?

haeter525 (changes requested):
Hi @gaoflow, thank you for the fix!
I have a few comments below for minor simplifications. Once those are addressed, I think the PR is ready to merge.

MickeyWzt:
I checked this PR locally on Windows and found one small portability issue in the current test suite.

At `28b0ef7`, running:

```bash
python -m pytest tests/test_cli.py -q
```

fails on Windows in `test_missing_rules_path_fails_when_rules_are_loaded`, because the error message uses `{rule!r}`. That escapes backslashes in Windows paths, so `str(missing_rules)` is not found in `result.output`.

Changing the message to use the plain path keeps the CLI output more readable and makes the assertion portable:

```python
f"Path {rule} does not exist."
```

I also tried the two requested simplifications locally:

- set `exists=False` directly on the `--rule` option
- extract the duplicated `Quark` mock setup into a pytest fixture

With those changes plus the message tweak:

```bash
python -m pytest tests/test_cli.py -q
# 3 passed

uv run --with black black --check tests/test_cli.py
# passed

git diff --check
# passed
```

binggao1230:
Thanks both. Addressed in efa3c54:

- Set `exists=False` directly on the `--rule` option and dropped the post-hoc param patch (@haeter525).
- Switched the error message to the plain path so the assertion is portable on Windows (@MickeyWzt).
- Extracted the duplicated Quark mock into a `mock_quark` fixture (@haeter525).

`pytest tests/test_cli.py` is green here (3 passed).

haeter525 (approved):
Thank you @gaoflow for the fix!

PR #944, opened 2026-06-30
[codex] fix rule path validation for custom summary rules

### LINKED ISSUE #936: `--rule` default path validated even when unused

MickeyWzt:
﻿## Summary

Fixes #936.

This delays validation of the default rules directory until the CLI actually needs to walk a rules directory. Commands that provide a specific JSON rule with `-s custom_rule.json` no longer fail during Click option parsing just because the default rules directory has not been downloaded yet.

## Root cause

The `--rule` option used `click.Path(exists=True)` with a default path. Click validates default values while parsing, so an absent default rules directory blocked commands even when the command used a specific rule file instead.

## Changes

- Remove eager `exists=True` validation from the `--rule` Click option.
- Add an explicit missing-directory check before walking the rules directory.
- Add a regression test that sets the default rule path to a missing directory and verifies a custom summary rule file is still accepted.

## Validation

- `python -m pytest tests/test_cli.py -q` passed.
- `python -m pytest tests/test_cli.py tests/test_freshquark.py -q` shows the new test passing, while the existing `tests/test_freshquark.py` tests fail on Windows path separator expectations (`/` vs `\\`), unrelated to this change.

### DIFF
--- a/quark/cli.py
+++ b/quark/cli.py
@@ -66,7 +66,7 @@ logo()
     "-r",
     "--rule",
     help="Rules directory",
-    type=click.Path(exists=True, file_okay=True, dir_okay=True),
+    type=click.Path(file_okay=True, dir_okay=True),
     default=f"{config.DIR_PATH}",
     required=False,
     show_default=True,
@@ -205,6 +205,12 @@ def entry_point(
 
         rule_path_list = [rule_filter]
     else:
+        if not os.path.exists(rule):
+            raise click.BadParameter(
+                "Rules directory does not exist.",
+                param_hint="--rule",
+            )
+
         rule_path_list = [
             os.path.join(dir_path, file)
             for dir_path, _, file_list in os.walk(rule)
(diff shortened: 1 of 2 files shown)


MickeyWzt:
﻿I added a second regression test for the missing `--rule` directory branch and verified the focused CLI tests locally:

```bash
python -m pytest tests/test_cli.py -q
# 2 passed
```

One note on the Codecov patch failure: this looks like a workflow checkout issue for fork PRs rather than a missing test. The `build` workflow runs on `pull_request_target`, and the Actions log for the previous run shows `actions/checkout` fetching and checking out the base commit `1df1d49...` from `origin/master`, while Codecov uploads that coverage report against the PR head commit. That means the uploaded coverage is generated from the base branch, not from this PR's changed files, so patch coverage reports the changed `quark/cli.py` lines as uncovered even though the new tests cover the added branch when run against the PR branch.

The latest push restores the source diff to the clean implementation and keeps the added regression tests focused on the behavior in this PR.

haeter525:
Hi @MickeyWzt, thank you for the contribution!

We're already working on this issue in #937, so I am going to close this PR to avoid duplication. We'd still love to have your input, so feel free to share ideas or leave a review on #937. Thanks again!

PR #945, opened 2026-07-01
fix: validate default rules path only when used

### LINKED ISSUE #936: `--rule` default path validated even when unused

qinfenyinluo:
## Summary

- Allow `--rule` to be parsed even when the default rules directory has not been downloaded yet.
- Report a clear warning only when Quark actually needs to load rules from the configured rules path.
- Add CLI regression tests for using a custom single rule file while the default rules path is missing.

## Root cause

`--rule` used `click.Path(exists=True)` together with the default rules directory. Click validated that default during argument parsing, so commands that supplied their own rule file via `--summary custom_rule.json` could fail before Quark reached the code path that uses the custom rule.

## Validation

- `python -m pytest tests\test_cli.py -q` -> 2 passed
- `python -m black --check tests\test_cli.py` -> passed
- `git diff --cached --check` -> passed before commit

I also tried `python -m pytest tests\test_cli.py tests\test_freshquark.py -q`. The new CLI tests passed, but the existing `tests/test_freshquark.py` assertions fail on Windows because expected paths use `\` while the mocked command builds paths with `/`; this appears unrelated to this change.

Closes #936

### DIFF
--- a/quark/cli.py
+++ b/quark/cli.py
@@ -66,7 +66,7 @@ logo()
     "-r",
     "--rule",
     help="Rules directory",
-    type=click.Path(exists=True, file_okay=True, dir_okay=True),
+    type=click.Path(exists=False, file_okay=True, dir_okay=True),
     default=f"{config.DIR_PATH}",
     required=False,
     show_default=True,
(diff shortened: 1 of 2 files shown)


qinfenyinluo:
Closing this as duplicate after noticing that #937 already covers the same issue. Sorry for the extra noise, and thanks for maintaining Quark.

PR #946, opened 2026-07-04
fix: defer default rule path validation (#936)

### LINKED ISSUE #936: `--rule` default path validated even when unused

gitcommit90:
## Summary
- Remove `exists=True` from the `--rule` option so Click does not reject the missing default path during parsing.
- Existing command logic still validates explicit JSON rule files and walks the rule directory.

Closes #936

### DIFF
--- a/quark/cli.py
+++ b/quark/cli.py
@@ -66,7 +66,7 @@ logo()
     "-r",
     "--rule",
     help="Rules directory",
-    type=click.Path(exists=True, file_okay=True, dir_okay=True),
+    type=click.Path(file_okay=True, dir_okay=True),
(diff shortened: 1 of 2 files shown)


haeter525:
Hi @gitcommit90. thank you for opening the PR!

We're already working on this issue in #937, so I am going to close this PR to avoid duplication. Please feel free to share ideas or leave a comment there. Thanks.

PR #953, opened 2026-07-13
fix: include interfaces in superclass relationships

### LINKED ISSUE #948: A line in apkinfo get_superclass_relationship does not have any effect
This line does not seem to have any effect https://github.com/ev-flow/quark-engine/blob/cf35a66292a185dc0585675c7fff3e5126d5ebf0/quark/core/apkinfo.py#L455-L457

`set.union` returns a new set, but does not modify original set in-place.

If it should be modified (like in method `subclass_relationship`), either elements should be `add`'ed, or `union` should be changed to `update`

wenxiaojie1:
## Summary
- include implemented interfaces in `superclass_relationships`
- add a regression test covering a class with a parent and two interfaces

## Validation
- `python -m compileall -q quark/core/apkinfo.py tests/core/test_apkinfo.py`
- `git diff --check`
- verified the expected set-update behavior with a minimal local fixture

The targeted pytest invocation could not be collected locally because this machine lacks the project's pinned `androguard==3.4.0a1` dependency; the new regression test will run in the repository CI environment.

### DIFF
--- a/quark/core/apkinfo.py
+++ b/quark/core/apkinfo.py
@@ -452,7 +452,7 @@ class AndroguardImp(BaseApkinfo):
 
         for _class in self.analysis.get_classes():
             hierarchy_dict[str(_class.name)].add(str(_class.extends))
-            hierarchy_dict[str(_class.name)].union(
+            hierarchy_dict[str(_class.name)].update(
                 str(implements) for implements in _class.implements
             )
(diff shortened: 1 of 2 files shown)


haeter525 (approved):
Hi @wenxiaojie1, thanks for the PR!

PR #855, opened 2026-01-18
Add DexTrace core-library backend

sidra-asa:
This PR adds the DexTrace-based apkinfo backend integration. 

## How to test

```Bash
$ git clone git@github.com:ev-flow/DexTrace.git
$ cd DexTrace
$ pip install -e .
$ quark -a APK -s --core-library dextrace

PR #958, opened 2026-08-01
Add DexTrace core-library backend (continuation of PR #855)

haeter525:
## Description

Adds `DexTraceImp` as a new concrete implementation of `BaseApkinfo`, allowing Quark to use DexTrace as an alternative APK analysis backend. Activate it with `--core-library dextrace`:

```bash
quark -a <apk_file> --core-library dextrace -s
```

## Key Changes

- `quark/core/dextraceapkinfo.py`: Add `DexTraceImp`, implementing the same `BaseApkinfo` interface as `AndroguardImp`.

- `quark/core/quark.py`:
  - Wire `core_library="dextrace"` to `DexTraceImp`.
  - Fix false positive in `find_api_usage` subclass BFS: check `class_name in current_class_set` instead of discarding `Ljava/lang/Object;` and testing non-empty.

- `tests/core/test_dextraceapkinfo.py`: Add integration tests covering:
  - Class name normalisation (Ahmyth APK).
  - Bogus ZIP compression and permissions (3d52b APK).
  - Manifest fast-fail (synthetic APK via `tmp_path`).

## Motivation and Context

DexTrace is a Python APK analysis library that parses DEX bytecode independently of the androguard package. Adding it as a backend gives Quark an alternative that does not depend on androguard. The implementation handles the anti-analysis patterns and class naming conventions seen in real-world malware samples.

## How Has This Been Tested?

- `pytest tests/` passes with no regressions.
- Full run on 497 APKs from the Quark malware family analysis report:
  - **Accuracy:** DexTrace's result is more accurate than Androguard's.
  - **Performance:** mean 27.5 s vs 5.7 s (4.8x faster), total wall time 57 min vs 11 min (5.2x faster).

### DIFF
--- a/quark/cli.py
+++ b/quark/cli.py
@@ -134,7 +134,7 @@ logo()
     "core_library",
     help="Specify the core library used to analyze an APK",
     type=click.Choice(
-        ("androguard", "rizin", "radare2", "shuriken"),
+        ("androguard", "rizin", "radare2", "shuriken", "dextrace"),
         case_sensitive=False
     ),
     required=False,
--- a/quark/core/dextraceapkinfo.py
+++ b/quark/core/dextraceapkinfo.py
@@ -1,0 +1,679 @@
+# -*- coding: utf-8 -*-
+# This file is part of Quark-Engine - https://github.com/ev-flow/quark-engine
+# See the file 'LICENSE' for copying permission.
+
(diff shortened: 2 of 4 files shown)


pulorsok (approved):
LGTM.

Issue #264, opened 2021-10-13
Error on GitHub actions Windows VM


Surendrajat:
**Describe the bug**

Quark is throwing error while running `quark --help`:

```
Traceback (most recent call last):
  File "c:\hostedtoolcache\windows\python\3.7.9\x64\lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "c:\hostedtoolcache\windows\python\3.7.9\x64\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\Scripts\quark.exe\__main__.py", line 4, in <module>
  File "c:\hostedtoolcache\windows\python\3.7.9\x64\lib\site-packages\quark\cli.py", line 13, in <module>
    from quark.core.parallelquark import ParallelQuark
  File "c:\hostedtoolcache\windows\python\3.7.9\x64\lib\site-packages\quark\core\parallelquark.py", line 8, in <module>
    from quark.core.quark import Quark
  File "c:\hostedtoolcache\windows\python\3.7.9\x64\lib\site-packages\quark\core\quark.py", line 12, in <module>
    from quark.core.apkinfo import AndroguardImp
  File "c:\hostedtoolcache\windows\python\3.7.9\x64\lib\site-packages\quark\core\apkinfo.py", line 20, in <module>
    class AndroguardImp(BaseApkinfo):
  File "c:\hostedtoolcache\windows\python\3.7.9\x64\lib\site-packages\quark\core\apkinfo.py", line 264, in AndroguardImp
    ) -> MethodObject:
  File "c:\hostedtoolcache\windows\python\3.7.9\x64\lib\functools.py", line 490, in lru_cache
    raise TypeError('Expected maxsize to be an integer or None')
TypeError: Expected maxsize to be an integer or None
```

[Complete logs](https://github.com/APKLab/APKLab/runs/3877955558#step:8:16)

**To Reproduce**
```bash
pip install -U quark-engine
freshquark
quark --help
```

**Expected behavior**
It should show the help.

**Desktop (please complete the following information):**
 - OS: Microsoft Windows Server 2019
 - Quark-engine v21.10.2

**Additional context**
It's running on GitHub actions, and I don't have a real Windows machine to test or fix this issue.

Issue #260, opened 2021-10-06
Version not update


sharteeya:
**Describe the bug**
I just found that the version in `__init__.py` does not update. I will push a fix for this very soon.

**Expected behavior**
The version should be 21.10.1.

**Screenshots**
![image](https://user-images.githubusercontent.com/38580996/136129130-dbeb34dd-2231-45ac-9e0a-6d6f33854f60.png)

PR #234, opened 2021-08-23
Fix pytest.xml

haeter525:
**Description**
1. Specify the Rizin version to the latest release, v0.2.1.
2. Cache the generated libraries after building Rizin.

**Code changes**
+ pytest.xml

PR #961, opened 2026-08-03
Add Quark malware analysis report for EventBot

pulorsok:
# EventBot Malware Family Analysis Report

This report analyses the [EventBot](https://malpedia.caad.fkie.fraunhofer.de/details/apk.eventbot) malware family using Quark's rule classification. EventBot is an Android banking trojan and information stealer, first documented by Cybereason in March 2020, that abuses Android's accessibility service to harvest data from more than 200 banking, money-transfer and cryptocurrency-wallet applications, and reads incoming SMS to defeat two-factor authentication. This run did not generate a new rule for EventBot: the family's distinctive bytecode patterns (installed-app enumeration, SMS broadcast interception, and SMS-body exfiltration over HTTP) are already covered by Quark's existing rule set. Check [here](https://github.com/quark-engine/quark-rules) for the rule set details.

Quark's rule classification flagged **3 of 3 EventBot samples** as high-risk in this experiment (detection rate **100%**). Benign-cohort false-positive rate was not measured here. See [tested APKs](#list-of-tested-apks) below.

This section uses MITRE ATT&CK Mobile as its reference taxonomy, anchored to the [S0478 EventBot](https://attack.mitre.org/software/S0478) software entry. Of the **14 techniques** documented for the family, Quark's static bytecode analysis surfaces the **3 listed below**; see the [coverage-gap notes](#coverage-gap-notes) at the end of this section for why the other 11 are not demonstrated here.

| MITRE Technique | Real-world manifestation in EventBot |
|---|---|
| T1418 Software Discovery | Enumerating installed applications and persisting the inventory to shared preferences, so the operator can pick which banking apps to target |
| T1624.001 Broadcast Receivers | Registering an SMS broadcast receiver that fires on every incoming message and hands the intent to a parser |
| T1636.004 SMS Messages | Reading the SMS body out of the received intent and shipping it to the C2 over HTTP, defeating SMS-based two-factor authentication |

All cluster representatives below were extracted from sample `05782e267bd62de78a3db22b1a83ddd3c72cbef95f5a5bc9defdd42a4f5786ec.apk`, chosen as the representative sample whose detected behaviors most fully cover the documented profile of EventBot. The other 2 family samples were used to compute the detection-rate figure above.

Each section below corresponds to one technique from the table above. Within each section we first quote the MITRE definition, then walk through the call sequence and list the underlying rules.

### 1. T1418 Software Discovery

[T1418 Software Discovery — attack.mitre.org](https://attack.mitre.org/techniques/T1418)

> **MITRE definition (T1418):** Adversaries may attempt to get a listing of applications that are installed on a device.

T1418 Software Discovery

`Lcom/example/eventbot/func;regOnWeb` — the routine that registers the freshly infected device with the operator's web panel — calls the helper `Lcom/example/eventbot/func;getAllApps`, which enumerates every installed application and writes the resulting list into shared preferences. The inventory is then included in the registration payload, letting the operator decide which of EventBot's 200-plus targeted banking apps are worth attacking on this device.

- Get installed applications and put the list in shared preferences (#00170)
- Enumerate installed applications (#00264)

### 2. T1624.001 Broadcast Receivers

[T1624.001 Broadcast Receivers — attack.mitre.org](https://attack.mitre.org/techniques/T1624/001)

> **MITRE definition (T1624.001):** Adversaries may establish persistence using system mechanisms that trigger execution based on specific events. Mobile operating systems have means to subscribe to events such as receiving an SMS message, device boot completion, or other device activities.

T1624.001 Broadcast Receivers

This section covers the *trigger* mechanism. `Lcom/example/eventbot/func;doSMS` is the broadcast-receiver entry point: it inspects the incoming intent's action to confirm an SMS-received broadcast, then dispatches to `Lcom/example/eventbot/func;parseSMS`. Because the receiver is bound to a system event rather than to app lifecycle, it runs whenever a message arrives, giving EventBot execution without the user opening the app.

- Monitor the general action to be performed (#00025)
- Monitor incoming SMS message (#00234)
- Read SMS message from Intents (#00260)

### 3. T1636.004 SMS Messages

[T1636.004 SMS Messages — attack.mitre.org](https://attack.mitre.org/techniques/T1636/004)

> **MITRE definition (T1636.004):** Adversaries may utilize standard operating system APIs to gather SMS messages. On Android, this can be accomplished using the SMS Content Provider. iOS provides no standard API to access SMS messages.

T1636.004 SMS Messages

This section covers what the receiver *collects*, once triggered by the mechanism in the previous section. `Lcom/example/eventbot/func;parseSMS` extracts the message body from the intent and passes it to EventBot's HTTP sender, which posts it to the C2. This is the step that defeats SMS-based two-factor authentication: one-time codes are read and exfiltrated before or alongside the victim seeing them.

- Monitor the general action to be performed (#00025)
- Get SMS message body and send it through http (#00165)
- Monitor incoming SMS message (#00234)
- Read SMS message from Intents (#00260)

### Coverage-gap notes

The 11 MITRE techniques documented for EventBot that this report does NOT demonstrate, grouped by the reason Quark could not confirm them on these samples:

**The rule's two APIs are both present, but no single method calls both (4)** — Quark confirms a rule only when one method reaches both of its APIs, so it can trace the value from the first call into the second. For each technique below a rule exists and its APIs are present in the samples (Quark's stage 2 passes), but EventBot splits the work across separate small helpers, so stage 3 never finds a method holding both.

| Technique | Rule that partially matched | Where it stopped |
|---|---|---|
| T1426 System Information Discovery | #00033 Query the IMEI number, #00038 Query the phone number | Both APIs present; no shared calling method |
| T1422 System Network Configuration Discovery | #00059 Query the SIM card status | Both APIs present; no shared calling method |
| T1422.001 Internet Connection Discovery | #00034 Query the current data network type | Both APIs present; no shared calling method |
| T1513 Screen Capture | screen-capture rules, matched in all three samples | Rules fired, but no method held enough of the chain to anchor a call sequence |

**Behavior confirmed, but every candidate representative sat in bundled library code (2)** — the evidence exists and passed validation, but the methods holding it belong to third-party code, so attributing the behavior to EventBot would be wrong.

| Technique | What was found | Why it was dropped |
|---|---|---|
| T1407 Download New Code at Runtime | Dynamic class loading plus method reflection, in the obfuscated sample | That sample's class names are ProGuard-minified, so its parents cannot be separated from shaded library code |
| T1437.001 Web Protocols | HTTP POST carrying the SMS body and sender address | The representative parent landed in the bundled HTTP stack. The underlying evidence still appears in this report — rule #00165 is cited under T1636.004 above |

**No rule in the current pool targets the technique (2)**

- **T1521.001 Symmetric Cryptography** and **T1406 Obfuscated Files or Information** — the rule pool currently contains no rule describing symmetric-encryption or payload-obfuscation behavior, so neither technique was exercised at all.

**The samples do not exhibit the behavior (3)** — these are honest non-detections rather than tooling limits. Cybereason documented EventBot as still under active development when discovered, and the capabilities below are absent from the builds in this cohort.

| Technique | What the samples actually do | What the technique requires |
|---|---|---|
| T1417.001 Keylogging | Accessibility UI inspection — `getRootInActiveWindow`, reading node text and content descriptions | An input-capture primitive; reading the view tree is not the same as recording keystrokes |
| T1417.002 GUI Input Capture | Passive dimension queries (navigation-bar and status-bar heights) | An overlay window or input interception |
| T1655.001 Match Legitimate Name or Location | Hiding the launcher icon | Evidence of impersonating a *specific* legitimate app by name, icon or package; icon hiding alone is generic |

## List of Tested APKs

| index | sha256 |
|-------|--------|
| 1 | 05782E267BD62DE78A3DB22B1A83DDD3C72CBEF95F5A5BC9DEFDD42A4F5786EC |
| 2 | 1A8C17AD1A790554278B055BDB946D4597BA9AF6BE3611EE6311B90C7F7848C5 |
| 3 | E6E736DF1923C47F6624DAB18BF6A341995B5E756395256AA0B3F4C43594481A |

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,7 @@
 | AhRat       | Android RAT capable of surveillance and data theft. | 1. Record audio from the device.<br>2. Steal files and sensitive data.<br>3. Remote access via C2.<br>4. Execute remote commands. | [View](https://quark-engine.readthedocs.io/en/latest/malware_report.html#ahrat-malware-family-analysis-report) |
 | AndroRat    | Android remote access trojan for device surveillance. | 1. Record audio and capture video.<br>2. Track device location.<br>3. Steal files and device information.<br>4. Execute remote commands. | [View](https://quark-engine.readthedocs.io/en/latest/malware_report.html#androrat-malware-family-analysis-report) |
 | Sova        | Android banking trojan distributed as trojanised carrier apps for credential theft and SMS fraud. | 1. Read device identifiers via the C2 ping-response handler.<br>2. Inject outbound SMS on operator command.<br>3. Place phone calls without user consent. | [View](https://quark-engine.readthedocs.io/en/latest/malware_report.html#sova-malware-family-analysis-report) |
+| EventBot    | Banking trojan and infostealer targeting 200+ financial apps. | 1. Enumerate installed applications to pick targets.<br>2. Intercept incoming SMS via a broadcast receiver.<br>3. Exfiltrate SMS bodies over HTTP to defeat 2FA. | [View](https://quark-engine.readthedocs.io/en/latest/malware_report.html#eventbot-malware-family-analysis-report) |
 
 ## Quick Start
 
--- a/docs/source/malware_report.rst
+++ b/docs/source/malware_report.rst
@@ -3318,3 +3318,167 @@ The table below lists the APKs we tested.
 +-------+------------------------------------------------------------------+
 | 14    | F3FC80A8793E60A901DA44B9AB315931699E64A4F3EDDB8ABA839FE860DE46DC |
 +-------+------------------------------------------------------------------+
+
+EventBot Malware Family Analysis Report
+=======================================
+
+This report analyses the `EventBot <https://malpedia.caad.fkie.fraunhofer.de/details/apk.eventbot>`__ malware family using Quark's rule classification. EventBot is an Android banking trojan and information stealer, first documented by Cybereason in March 2020, that abuses Android's accessibility service to harvest data from more than 200 banking, money-transfer and cryptocurrency-wallet applications, and reads incoming SMS to defeat two-factor authentication. This run did not generate a new rule for EventBot: the family's distinctive bytecode patterns (installed-app enumeration, SMS broadcast interception, and SMS-body exfiltration over HTTP) are already covered by Quark's existing rule set. Check `here <https://github.com/quark-engine/quark-rules>`__ for the rule set details.
+
+Quark's rule classification flagged **3 of 3 EventBot samples** as high-risk in this experiment (detection rate **100%**). Benign-cohort false-positive rate was not measured here. Please check :ref:`here <list-of-tested-apks-eventbot>` for the APKs we tested.
+
+Identified Well-Known Threats
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This section uses MITRE ATT&CK Mobile as its reference taxonomy, anchored to the `S0478 EventBot <https://attack.mitre.org/software/S0478>`__ software entry. Of the **14 techniques** documented for the family, Quark's static bytecode analysis surfaces the **3 listed below**; see the :ref:`coverage-gap notes <coverage-gap-notes-eventbot>` at the end of this section for why the other 11 are not demonstrated here.
+
+.. list-table::
+   :header-rows: 1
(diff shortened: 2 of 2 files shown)


haeter525 (approved):
LGTM. Thanks.

PR #968, opened 2026-08-25
Add Quark malware analysis report for Chrysaor

pulorsok:
# Chrysaor Malware Family Analysis Report

This report analyses the [Chrysaor](https://malpedia.caad.fkie.fraunhofer.de/details/apk.chrysaor) malware family using Quark's rule classification. Chrysaor is the Android build of the Pegasus spyware developed by NSO Group, documented by Lookout and Google in 2017. It is delivered as a targeted implant that records audio, reads stored messages, and can remove its own traces on command. This run did not add a new rule for Chrysaor. Check [here](https://github.com/quark-engine/quark-rules) for the rule set details.

Quark's rule classification flagged **2 of 2 Chrysaor samples** as high-risk in this experiment (detection rate **100%**). Benign-cohort false-positive rate was not measured here. See [tested APKs](#list-of-tested-apks) below.

This section uses MITRE ATT&CK Mobile as its reference taxonomy, anchored to the [S0316 Pegasus for Android](https://attack.mitre.org/software/S0316) software entry — Chrysaor is MITRE's alias for the same implant. Of the **14 techniques** documented for the family, Quark's static bytecode analysis confirms the **2 listed below**; see the [coverage-gap notes](#coverage-gap-notes) at the end of this section for why the other 12 are not demonstrated here.

| MITRE Technique | Real-world manifestation |
|---|---|
| T1429 Audio Capture | Recording ambient audio and phone calls for surveillance and intelligence gathering |
| T1422.001 Internet Connection Discovery | Checking whether the device currently has an active network connection |

All behavior maps below were rendered from sample `ade8bef0ac29fa363fc9afd958af0074478aef650adeb0318517b48bd996d5d5.apk` (package `com.network.android`) — chosen as the representative sample whose detected behaviors most fully cover the documented profile of Chrysaor. The other family sample was used to compute the detection-rate figure above.

Each section below corresponds to one technique from the table above. Within each section we first quote the MITRE definition, then show the Quark behavior map extracted from the representative sample's bytecode, then walk through the call sequence and list the underlying rules.

### 1. T1429 Audio Capture

[T1429 Audio Capture — attack.mitre.org](https://attack.mitre.org/techniques/T1429)

> **MITRE definition (T1429):** Adversaries may capture audio to collect information by leveraging standard operating system APIs of a mobile device. Examples of audio information adversaries may target include user conversations, surroundings, phone calls, or other sensitive information.

T1429 Audio Capture

`Lcom/network/android/roomTap/AutoAnswerReceiver;c` directly initializes the recorder and starts recording. This call enables the receiver to capture ambient audio without user interaction.

- Initialize the recorder and start recording (#00198)

### 2. T1422.001 Internet Connection Discovery

[T1422.001 Internet Connection Discovery — attack.mitre.org](https://attack.mitre.org/techniques/T1422/001)

> **MITRE definition (T1422.001):** Adversaries may check for Internet connectivity on compromised systems.

T1422.001 Internet Connection Discovery

`Lcom/network/f/a;a` directly checks for network connectivity. This call enables the malware to determine whether an active network connection is available.

- Check for network connectivity (#00224)

### Coverage-gap notes

The 12 MITRE techniques documented for Chrysaor that this report does NOT demonstrate, grouped by the reason Quark could not confirm them:

**What makes it malicious is in the text, not the call (2)** — T1404 Exploitation for Privilege Escalation and T1636.002 Call Log. `Lcom/network/android/m` runs shell commands and Quark sees the call, but a shell call looks the same whatever it runs. T1636.002 fails the same way: the rules that fired cover the SMS and call-log providers together, because both are reached through one API with a different URI string.

**The samples do not exhibit the behaviour (7)** — T1409 Stored Application Data, T1418 Software Discovery, T1422 System Network Configuration Discovery, T1422.002 Wi-Fi Discovery, T1636.001 Calendar Entries, T1644 Out of Band Data and T1645 Compromise Client Software Binary. These samples do not carry these behaviours, so there is nothing for Quark to detect.

**The two calls happen but nothing is passed between them (3)** — T1512 Video Capture, T1624.001 Broadcast Receivers and T1636.003 Contact List. Each reaches stage 4: one method makes both calls, in the right order. The camera object, for one, travels through wrapper methods instead of going straight from the first call to the second, so Quark cannot link them.

## List of Tested APKs

| index | sha256 |
|-------|--------|
| 1 | ADE8BEF0AC29FA363FC9AFD958AF0074478AEF650ADEB0318517B48BD996D5D5 |
| 2 | BD8CDA80AAEE3E4A17E9967A1C062AC5C8E4AEFD7EAA3362F54044C2C94DB52A |

---

### DIFF
--- a/docs/source/malware_report.rst
+++ b/docs/source/malware_report.rst
@@ -3606,3 +3606,85 @@ The table below lists the APKs we tested.
 +-------+------------------------------------------------------------------+
 | 2     | E6ABA7629608A525B020F4E76E4694D6D478DD9561D934813004B6903D66E44C |
 +-------+------------------------------------------------------------------+
+
+Chrysaor Malware Family Analysis Report
+=======================================
+
+This report analyses the `Chrysaor <https://malpedia.caad.fkie.fraunhofer.de/details/apk.chrysaor>`__ malware family using Quark's rule classification. Chrysaor is the Android build of the Pegasus spyware developed by NSO Group, documented by Lookout and Google in 2017. It is delivered as a targeted implant that records audio, reads stored messages, and can remove its own traces on command. This run did not add a new rule for Chrysaor. Check `here <https://github.com/quark-engine/quark-rules>`__ for the rule set details.
+
+Quark's rule classification flagged **2 of 2 Chrysaor samples** as high-risk in this experiment (detection rate **100%**). Benign-cohort false-positive rate was not measured here. Please check :ref:`here <list-of-tested-apks-chrysaor>` for the APKs we tested.
+
+Identified Well-Known Threats
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The behaviours below are referenced from `MITRE ATT&CK® Mobile — S0316 Pegasus for Android <https://attack.mitre.org/software/S0316>`__.
+
+This section uses MITRE ATT&CK Mobile as its reference taxonomy, anchored to the `S0316 Pegasus for Android <https://attack.mitre.org/software/S0316>`__ software entry — Chrysaor is MITRE's alias for the same implant. Of the **14 techniques** documented for the family, Quark's static bytecode analysis confirms the **2 listed below**; see the coverage-gap notes at the end of this section for why the other 12 are not demonstrated here.
+
+.. list-table::
+   :header-rows: 1
+   :widths: 30 70
+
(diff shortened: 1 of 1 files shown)


haeter525 (approved):
LGTM

Issue #279, opened 2021-10-27
Different JSON formats cause Jadx and APKLab to fail to read Quark report


krnick:
**Describe the bug**
In the feature request in issue #117, there are two PRs (#254, #247) related to it and modified the original report format, including the detail report and JSON report, which caused Jadx and APKLab to fail to read different JSON formats.

**Quark-Engine:**
 - Version v21.10.2

Issue #117, opened 2021-02-09
Unable to get function involved in a 60% confident rule


cryptax:
For rules less than 100% confident, I would like to view **which portion of code is concerned**.
This _works very well with rules detected at 100%_ and option `-c`. I get the functions of the malware concerned by this or that crime. Example:

![quark-c1](https://user-images.githubusercontent.com/2493690/107363948-20773f00-6adb-11eb-894c-305bc80e254a.png)

But **I am unable to get the same result with rules less than 100%.** I am using this [malicious sample](https://koodous.com/apks/a03ffa955d9b273cc22acae93dbad2ec8c023d7621b256ea93506587506e3eb3). 
On my sample, Quark detects rule `00070.json` at 60% .

I tried: `quark -s -t 20 -c -a a03ffa955d9b273cc22acae93dbad2ec8c023d7621b256ea93506587506e3eb3.apk -r ./test-rules/`. Despite the `-c`, and `-t 20` to consider rules above 20%, I do not see the function involved.See image below: there is nothing underneath "Rules Classification".

![quark-bug-class](https://user-images.githubusercontent.com/2493690/107362724-6cc17f80-6ad9-11eb-822e-1dc0551dc71f.png)

I also tried with `-d` instead of `-s`. Still do not get the function involved.

![quark-bug2](https://user-images.githubusercontent.com/2493690/107362806-8a8ee480-6ad9-11eb-8f33-f82ebca37aa0.png)

 I tried with `-g` to generate the call graph, which should give me the information. 

![quark-bug3](https://user-images.githubusercontent.com/2493690/107362967-c2962780-6ad9-11eb-8cee-58690b897c52.png)

It says it generates the graph, but I am unable to find it. I did a `ls -lt` on the directory nothing recent, apart from a file `rules_classification.json` which contains:

```
{"rules_classification": []}
```
I suspect the issue is around this: for some reason, my rule does not get selected even the threshold is set, so therefore there is nothing to generate in the call graph.

Finally, I tried with an output report `-o` : `quark -a a03ffa955d9b273cc22acae93dbad2ec8c023d7621b256ea93506587506e3eb3.apk -r ./test-rules/ -s -c -o ./quark.output`
The output report says the following (but still unfortunately does not show the functions involved):

```json
{
    "md5": "aaae379a27d5355231fcefb776f2df84",
    "apk_filename": "a03ffa955d9b273cc22acae93dbad2ec8c023d7621b256ea93506587506e3eb3.apk",
[...]

Issue #11, opened 2019-11-20
Welcome gwalstat


krnick:
gwalstat is a robot for typo checking

krnick:
Thanks for the report @krnick!                         I will look into it ASAP! (I'm a bot).

18z:
Good to see you here, gwalstat!

PR #976, opened 2026-09-22
Add Quark malware analysis report for Exodus

pulorsok:
# Exodus Malware Family Analysis Report

This report analyses the [Exodus](https://malpedia.caad.fkie.fraunhofer.de/details/apk.exodus) malware family using Quark's rule classification. Exodus is Android surveillanceware written by the Italian firm eSurv and documented by Security Without Borders in 2019. It was delivered through decoy apps posing as mobile-operator service tools, and installs in two stages: a first stage that registers the device with the operator's server, and a surveillance implant downloaded afterwards. The three samples analysed here are the first stage. This run added a new rule (#00279), which flags reading the device IMEI and the subscriber IMSI in the same method; it is demonstrated in the T1422 section below. Check [here](https://github.com/quark-engine/quark-rules) for the rule details.

Quark's rule classification flagged **3 of 3 Exodus samples** as high-risk in this experiment (detection rate **100%**). Benign-cohort false-positive rate was not measured here. See [tested APKs](#list-of-tested-apks) below.

This section uses MITRE ATT&CK Mobile as its reference taxonomy, anchored to the [S0405 Exodus](https://attack.mitre.org/software/S0405) software entry. Of the **19 techniques** documented for the family, Quark's static bytecode analysis confirms the **3 listed below**; see the [coverage-gap notes](#coverage-gap-notes) at the end of this section for the other 16.

| MITRE Technique | Real-world manifestation |
|---|---|
| T1418 Software Discovery | Listing the installed applications to check whether a given package is present on the device |
| T1422 System Network Configuration Discovery | Collecting the device IMEI and the subscriber IMSI and sending them to the operator's server |
| T1636.002 Call Log | Reading the device's call-log provider |

All behavior maps below were rendered from sample `4f6146956b50ae3a6e80a1c1f771dba848ba677064eb0e166df5804ac2766898.apk` (package `operatore.italia`, posing as an Italian operator's contact-sync app) — chosen as the representative sample whose detected behaviors most fully cover the documented profile of Exodus. The other 2 family samples were used to compute the detection-rate figure above.

Each section below corresponds to one technique from the table above. Within each section we first quote the MITRE definition, then show the Quark behavior map extracted from the representative sample's bytecode, then walk through the call sequence and list the underlying rules.

### 1. T1418 Software Discovery

[T1418 Software Discovery — attack.mitre.org](https://attack.mitre.org/techniques/T1418)

> **MITRE definition (T1418):** Adversaries may attempt to get a listing of applications that are installed on a device.

T1418 Software Discovery

`Llib/operatore/core/a;a` walks the full `getInstalledApplications` list and compares each package name against the one it is looking for. Its caller uses that answer to decide whether to start the registration routine.

- Enumerate installed applications (#00264)

### 2. T1422 System Network Configuration Discovery

[T1422 System Network Configuration Discovery — attack.mitre.org](https://attack.mitre.org/techniques/T1422)

> **MITRE definition (T1422):** Adversaries may look for details about the network configuration and settings, such as IP and/or MAC addresses, of devices they access or through information discovery of remote systems. On Android, details of onboard network interfaces are accessible to apps through the `java.net.NetworkInterface` class. Previously, the Android `TelephonyManager` class could be used to gather telephony-related device identifiers, information such as the IMSI, IMEI, and phone number.

T1422 System Network Configuration Discovery

`Llib/operatore/core/b;c` reads the device IMEI and the subscriber IMSI in one method, along with the network operator name. This is the method that registers the device with the server, and the new rule added by this run fires on the IMEI-and-IMSI pair.

- Get the IMSI and network operator name (#00117)
- Collect device IMEI and subscriber IMSI identifiers (#00279)

### 3. T1636.002 Call Log

[T1636.002 Call Log — attack.mitre.org](https://attack.mitre.org/techniques/T1636/002)

> **MITRE definition (T1636.002):** Adversaries may utilize standard operating system APIs to gather call log data. On Android, this can be accomplished using the Call Log Content Provider. iOS provides no standard API to access the call log.

T1636.002 Call Log

`Loperatore/italia/ProgressActivity$2;a` queries the call-log content provider while the decoy "optimisation" screen runs. It takes the number of rows from the cursor; the call records themselves are not read out here.

- Query data from URI (SMS, CALLLOGS) (#00011)

### Coverage-gap notes

The 16 MITRE techniques documented for Exodus that this report does NOT demonstrate, grouped by the reason Quark could not confirm them on these samples:

**The behaviour belongs to the second stage, which these samples only download (10)** — T1421 System Network Connections Discovery, T1422.001 Internet Connection Discovery, T1429 Audio Capture, T1430 Location Tracking, T1512 Video Capture, T1513 Screen Capture, T1532 Archive Collected Data, T1636.001 Calendar Entries, T1636.003 Contact List and T1636.004 SMS Messages. Rules for location, screen capture, calendar reads, SMS reads and network-state checks do fire in these APKs, but every one of them lands in the bundled AndroidX support or logback libraries, and none of those methods is reachable from Exodus's own code. The location hit is AppCompat's TwilightManager deciding whether to apply the night theme; the SMS-and-call-log hit is a support-library helper that reads a column from whatever content URI it is handed. Neither result reaches Exodus.

**Quark reached Exodus's own code, but the primitive is not this technique (3)** — T1404 Exploitation for Privilege Escalation, T1409 Stored Application Data and T1533 Data from Local System. The only `Runtime.exec` calls in Exodus's own code run `chmod 700` over the files it has just unpacked, which makes the downloaded payload executable rather than exploiting anything. The file reads that matched T1409 and T1533 come from that same routine — Exodus reading back its own downloaded payload, not other applications' data. Quark reports all of it under one name, `Llib/operatore/core/f;a`, because three different methods in that class are called `a`.

**The behaviour is real, but no rule covers the API it uses (3)** — T1407 Download New Code at Runtime, T1437.001 Web Protocols and T1509 Non-Standard Port. Exodus downloads its second stage and talks to its server through OkHttp; the pool's HTTP rules match `HttpURLConnection` and Apache HttpClient, and none match OkHttp. This run did generate a candidate rule for the step that follows the download (`Context.getFilesDir` paired with `Runtime.exec`), but it trained to a negative score — it also fires on benign apps — so it was not added. For T1509 there is nothing non-standard to find: the server is reached over HTTPS on port 443.

## List of Tested APKs

| index | sha256 |
|-------|--------|
| 1 | 0F5F1409B1EBBEE4AA837D20479732E11399D37F05B47B5359DC53A4001314E5 |
| 2 | 26FEF238028EE4B5B8DA631C77BFB44ADA3D5DB8129C45DEA5DF6A51C9EA5F55 |
| 3 | 4F6146956B50AE3A6E80A1C1F771DBA848BA677064EB0E166DF5804AC2766898 |


haeter525 (approved):
LGTM.

